PR #6, opened 2017-02-03
Close the menu when the query becomes empty

tvararu:
This was introduced in #5.

Also clean up some unnecessary lines from a few tests.


src/typeahead.jsx
```diff
@@ -96,6 +96,8 @@ export default class Typeahead extends Component {
           options
         })
       })
+    } else if (queryEmpty) {
```
danmitchell-: Good catch 💪

PR #10, opened 2017-02-16
Accessibility hardening

tvararu:
I've done some fixes for a variety of browser and assistive technology combinations.

Commits have info about every individual fix.


src/typeahead.jsx
```diff
@@ -224,10 +234,12 @@ export default class Typeahead extends Component {
 
     const Option = ({ children, idx }) =>
       <li
+        aria-selected={selected === idx}
```
danmitchell-: Nice 👍

PR #20, opened 2017-03-09
First pass of styling tweaks

edwardhorsford:
Styling updates for the picker ahead of user research.

Changes:

- Introduce concept of 'inline' vs 'overlay' - where the picker inserts itself in the page rather than on top of the page. Configurable with a class
- Change line weights
- Change text colours (with alternate shading goes as low as 11.1:1
- Add alternate background shading between options.

Before:
screen shot 2017-03-09 at 10 51 37

After:
screen shot 2017-03-09 at 10 51 45


tvararu:
This looks good. 👍 

Do you mind if I rebase it and clean up some bits? Primarily I want to move more of the inline styles to the stylesheet.

edwardhorsford:
Do whatever you like!

PR #21, opened 2017-03-09
No options content

edwardhorsford:
Update content and styling for "no options found"

Before:
screen shot 2017-03-09 at 12 25 13

After:
screen shot 2017-03-09 at 12 24 59

@tvararu ideally this content would be configurable.


tvararu:
👌

PR #24, opened 2017-03-15
Fix minor syntax error in README example

richardTowers:
There should be a comma between the element and id object fields.


tvararu:
Thanks. 👍

PR #26, opened 2017-03-15
Fix documentation

36degrees:
Fix invalid stylesheet and incorrect whitespace breaking headings


tvararu:
👌  I was wondering what happened to those.

PR #27, opened 2017-03-28
Add `enhanceSelectElement` function and `onSelect` property

### LINKED ISSUE #25: Support progressive enhancement of selects for users without JS
GOV.UK Verify are currently using https://github.com/JamieAppleseed/selectToAutocomplete in an alpha product. This is a jQuery plugin that takes a `select` tag as input and auto-magically transforms it into a typeahead.

The main advantage of this approach is that it makes the no-javascript fallback much easier to implement - you simply stamp out your `select` and `option` tags, and if the JS fails to run that's Good Enough(TM).

Of course it would still be possible to put a `select` in a `noscript` tag (or similar) to provide a no-javascript fallback while using `accesible-typeahead`, but the "provide a select and the plugin does the rest" is a nice workflow.

If you think that would be a nice feature I'd be happy to have a play and send you a PR. What do you think?

tvararu:
This PR adds a `AccessibleTypeahead.enhanceSelectElement` helper function which will:

- Place a typeahead input field adjacent to the specified `<select>`
- Set the typeahead `defaultValue` to the `option[selected]` if any
- Default the typeahead `id` to the `<select>`'s `id`
- Default the typeahead `name` attribute to `''` to prevent it being included in form submissions
- Default the typeahead `source` to a basic one that uses any existing `<option>`s from the `<select>`
- Hide the `<select>` using inline `display: none`
- Set the `<select>`'s id to `${id}-select` to decouple from any `<label>`
- Upon selecting a value in the typeahead, update the original `<select>`

Also includes some README reorg/cleanup.

Closes #25.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 _(add items here for easier creation of next log entry)_
 
+- Add `AccessibleTypeahead.enhanceSelectElement` function.
+- Add `onSelect` property.
+
 ## 0.3.1 - 2017-03-09
 
 - Add ability to specify a `defaultValue` to prefill the input.
--- a/README.md
+++ b/README.md
@@ -118,6 +118,12 @@ The minimum number of characters that should be entered before the typeahead wil
 
 The `name` for the typeahead input field, to use with a parent `<form>`.
 
+#### `onSelect: Function` (default: `() => {}`)
+
+Arguments: `query: string`
+
+This function will be called when the user selects an option, with the option they've selected.
+
 ## Progressive enhancement
 
 If your typeahead is meant to select from a small list of options (a few hundreds), we strongly suggest that you render a `<select>` menu on the server, and use progressive enhancement.
--- a/README.md
+++ b/README.md
@@ -56,41 +56,13 @@ AccessibleTypeahead({
 
 > :warning: WARNING: This is a work in progress and will change significantly. :warning:
 
-### `autoselect: boolean` (optional, default: `false`)
+### Required options
 
-Set to true to highlight the first option when the user types in something and receives results. Pressing enter will select it.
-
-### `cssNamespace: String` (optional, default: `'typeahead'`)
-
-The default CSS classes use [BEM](http://getbem.com/) with `typeahead` as the block name. If you already have CSS rules for `.typeahead--menu` or any of the other default classes, you can use this property to rename them and prevent clashes.
-
-TODO: Better styling docs.
-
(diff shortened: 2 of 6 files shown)


igneosaur:
Hi @tvararu, this enhanced select option is causing an error on the WAVE evaluation tool. It's because the `select` now has no label associated with it because the ID is changed as part of the `enhanceSelectElement` function. I understand why this is happening because the new `input` takes over but the select is still present on the page but it still retains the `name` attribute.

My question is: is there any reason the `<select>` element needs to remain on the page or can we maybe remove it after enhancement?

tvararu:
@igneosaur `enhanceSelectElement` is one way to solve the problem of making select elements easier to use, but not the only way.

This implementation hides the `<select>` element visually and then introduces an autocomplete that, behind the scenes, updates the content of the select.

Because the select retains its original `name=`, this means that a form that relies on a `<select>` initially can be progressively enhanced to use autocomplete functionality. The extra `<input name="text">`s are ignored by the server.

Some upsides to this are:

- It's progressively enhanced, and you get a usable select without JS
- It works irrespective of your backend and requires no changes in form handling. You're still sending the original select's items
- If the values of the select items are things like `gb` but visually they look like "Great Britain" to the user, that's something that can be done easily with a `<select>` whereas an `<input type="text">` isn't as trivial

Some downsides:

- If the user types in gibberish, nothing will be selected in the `<select>`. Or potentially, the wrong thing will stay selected. The basic implementation never addresed this, and there are some issues floating around about it. This is a pretty serious usability issue
- Wave and automated accessibility tools point out that the `<select>` has been rendered inaccessible. But this is a false positive/red herring, as we're intentionally enhancing it to a different interface, the autocomplete. It's safe to ignore these, but they do get flagged

To answer if the `<select>` *needs* to remain on the page, no! You can add bespoke logic to remove it after enhancement, and if you set the `name` attribute on the autocomplete, your backend will then get the value submitted by the user in the autocomplete. Whether this works well depends on your specific app and usecase; at the least, you'll need some error handling for situations where users type in gibberish (which they can't do as easily with the `<select>`, though it's still possible my manipulating the DOM elements).

Hope that background helps! 👍

igneosaur:
Thanks for your reply @tvararu. Yes, this is exactly what I ended up doing; removing the select and passing in the name of the original `selectElement`.

I wonder if this WAVE error is something you'd like flagging? Happy to get stuck in here with solutions too. But yeh, I suppose dealing with this in implementation is fair enough.

PR #32, opened 2017-04-03
Add accessibility-criteria.md

tvararu:
This describes the necessary behaviours that a typeahead needs to meet to be usable by assistive technologies.

In the future we might want to move these to more prominent documentation, or enhance them with automated tests that verify the outlined behaviours.

### DIFF
--- a/accessibility-criteria.md
+++ b/accessibility-criteria.md
@@ -1,0 +1,31 @@
+# Accessibility Acceptance Criteria
+
+Authors: [Theodor Vararu](https://github.com/tvararu), [Léonie Watson](https://github.com/LJWatson).
+
+## What's this?
+
+This describes the necessary behaviours that a typeahead (referred in this document as an "autocomplete") needs to meet to be usable by assistive technologies.
+
+They are useful if you are evaluating `accessible-typeahead`, or a different autocomplete solution.
+
+## User story
+
+> As an Assistive Technology (AT) user, I want to be able to search using autocomplete, so I can find and choose a matching result easily and accurately.
+
+## Acceptance criteria
+
+The field with autocomplete must:
+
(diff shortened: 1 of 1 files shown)


edwardhorsford:
I'm not sure about the language of 'selected' and 'confirmed'. Navigate and selected?

I'm also not entirely convinced that telling the user the number of matches is a _must_.

Possible additional ones:
* Inform the user when a match is confirmed.
* Inform the user if a match is pre-confirmed.

tvararu:
@edwardhorsford I've updated these. Can you give them another look?

edwardhorsford:
Looks better! 

Others that spring to mind:
* Inform the user if there's a default answer.
* Inform the user if there are no matches.

tvararu:
> Inform the user if there's a default answer.

We have "Inform the user when a match is confirmed" which should mean the same thing?

> Inform the user if there are no matches.

I think this is a dupe of: "Inform the user that N number of matches have been displayed"

edwardhorsford:
To clarify my points - the first was about telling users on page load if there's already a result pre-populated (they've returned from a check your answers page).

For the second, I agree with you - we don't need to do anything else.

---

I wonder whether the typeahead needs to let users know whether they *must* select from the list or whether they can submit anything in the textfield and the results are just there to be helpful.

tvararu:
> I wonder whether the typeahead needs to let users know whether they must select from the list or whether they can submit anything in the textfield and the results are just there to be helpful.

I think this ties into accessibility of the *entire form* and is potentially out of scope for the typeahead. However, thinking about that [made me open this issue](https://github.com/alphagov/accessible-typeahead/issues/59).

Thanks for raising all the good points, @edwardhorsford. 👍

PR #36, opened 2017-04-05
Add the `placeholder` option

tvararu:
This option will populate the placeholder attribute on the input element.


tvararu:
Thanks @danmitchell-!

PR #40, opened 2017-04-06
Don't focus suggestions when hovering them

tvararu:
The way hovering previously worked is it would `.focus` the underlying element, which would apply the focus styles.

The reason it was done this way was because the component was built with keyboard in mind before mouse; for keyboard use to work well with assistive technologies, focusing is necessary for the current implementation. When I added mouse support, I made the mouse perform the same action as the keyboard, which is focusing.

This had an added benefit: when using a screenreader like VoiceOver, focusing on suggestions while the user mouses over them would make VoiceOver speak them out, which could be seen as a win for users with enough sight to use the mouse, but that need to rely on a screen reader to actually read words.

However, the side effect to this is when using the typeahead with lots of options. Focusing while you're scrolling is jittery, because focusing an element inside a scrollable parent will always attempt to scroll it into view.

This is a breaking change because it changes existing behaviour with assistive technologies and also the public CSS API.

This is done using a JavaScript-based `--hovered` class instead of a CSS `:hover` pseudo-class. This is because when the user has focused an element using the keyboard, and uses the mouse to hover over something else, we want the mouse hover to take priority and not highlight the focused element. This is not possible using CSS because there [it doesn't have a no sibling selector](http://stackoverflow.com/questions/7866784/css-no-sibling-selector).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 _(add items here for easier creation of next log entry)_
 
+- [Breaking] Don't focus suggestions when hovering them, add `:hover` CSS class.
 - Add `showNoOptionsFound` property to allow users to disable this behaviour.
 - Pass through unrecognised key events to input, allowing users to continue typing when they are focusing an option.
 
--- a/examples/styled.css
+++ b/examples/styled.css
@@ -82,7 +82,8 @@
   background-color: #FAFAFA;
 }
 
-.typeahead__option--focused {
+.typeahead__option--focused,
+.typeahead__option:hover {
   background-color: #005EA5;
   border-color: #005EA5;
   color: white;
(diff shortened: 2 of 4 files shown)


tvararu:
Thanks for the code review @danmitchell-!

@edwardhorsford could you read the problem description and point out if I missed anything?

tvararu:
Added "Do not merge" because I'm not happy with the performance of the `--hovered` class being applied. I'd prefer to change to:

- Still keep track of which element is hovered using `state.hovered` and use it to unapply the `--focused` class
- Don't create a `--hovered` modifier class
- Tie the CSS to the `:hover` state as before

This creates an incongruence/break of contract between `state.hovered` and the `:hover` state but I think in this situation it's justified.

Issue #39, opened 2017-04-06
Update examples to include more countries


tvararu:
The current examples [only define 3 countries](https://github.com/alphagov/accessible-typeahead/blob/master/examples/index.html#L107-L109). Having at least one example with lots of countries would help with testing scrolling behaviours.

ganes1410:
I will do this.How many countries do i need to add?

tvararu:
Hi @ganes1410, feel free to hardcode a list based on something like the [GOV.UK country register](https://country.register.gov.uk/records) 👍

GeneralMeow:
Hey there,
is this still open to work on?

tvararu:
Closed by #45.

tvararu:
@GeneralMeow sorry! This one's been done. Please pick up another issue or come back in a while after more pop up. :)

PR #45, opened 2017-04-07
Added new Countries

ganes1410:
Added new Countries as discussed on #39 .Sorry about closing last two pull requests,I'm new to this.


tvararu:
Thanks @ganes1410! This is what I asked for, but can I ask for a few changes?

- I recognised I said "a few countries" but I think it would be best to just include all of them.
- Could you sort them in alphabetical order? (That means moving the previous entries too)
- (This is very minor) Our projects follow [git commit guidelines](https://github.com/alphagov/styleguides/blob/master/git.md), could you change the commit message to the present tense? And you can also add a bit more information about where the countries were picked from, if you used the [country register](https://country.register.gov.uk/records).

ganes1410:
@tvararu  I changed the things you asked for and [this](https://en.wikipedia.org/wiki/List_of_national_capitals_in_alphabetical_order) is where i pciked the countries from.

examples/index.html
```diff
@@ -104,10 +104,257 @@
...
+  "Uzbekistan",
+  "Vanuatu",
+  "Vatican City",
+  "Venezuela",
+  "Vietnam",
+  "Wallis and Futuna",
+  "Yemen",
+  "Zambia",
+  "Zimbabwe"
+]
```
tvararu: Can you indent these to the right so they line up as they did before, and use single quotes for the strings? It should look like this:

```js
        var results = [
          'Abkhazia',
          'Afghanistan',
          'Akrotiri and Dhekelia',
          ...
          'Zambia',
          'Zimbabwe'
        ]
```

examples/index.html
```diff
@@ -104,257 +104,263 @@
...
-  "Vanuatu",
-  "Vatican City",
-  "Venezuela",
-  "Vietnam",
-  "Wallis and Futuna",
-  "Yemen",
-  "Zambia",
-  "Zimbabwe"
-]
+          "Afghanistan",
```
ganes1410: is it okay now ?

examples/index.html
```diff
@@ -104,10 +104,263 @@
     <script type="text/javascript">
       function suggest (query, syncResults) {
         var results = [
-          'France',
-          'Germany',
-          'United Kingdom'
-        ]
+          "Afghanistan",
```
tvararu: Can you change the double quotes to single quotes?

examples/index.html
ganes1410: @tvararu  fixed all.But made many mistakes that's why this many commits

ganes1410:
@tvararu  can i write commit messages this long?

tvararu:
> @tvararu can i write commit messages this long?

@ganes1410 You sure can! How to do it depends on which way you are writing the commits.

If you are using an editor like Vim or Sublime Text, you can separate the parts of the commit message with double carriage returns, like this:

```
Header

Body

Body
```

And so on. You can write as much as you want.

If you are using the command line interface, you can specify separate `-m` sections for every part:

```bash
git commit -m "Header" -m "Body" -m "Body"
```

To write the commit I illustrated above, you can write:

```bash
git commit -m "Add full list of countries to examples" -m "The list is taken from https://en.wikipedia.org/wiki/List_of_national_capitals_in_alphabetical_order"
```

But [using a git editor](http://stackoverflow.com/questions/2596805/how-do-i-make-git-use-the-editor-of-my-choice-for-commits) is probably easier, and it's how I do it.

examples/index.html
```diff
@@ -162,7 +162,7 @@
           'Cook Islands',
           'Coral Sea Islands',
           'Costa Rica',
-          'Cote d'Ivoire',
+          'Cote d\'Ivoire',
```
tvararu: Good catch.

ganes1410:
@tvararu  is everything ok now ??

tvararu:
@ganes1410 yes! Thank you. 👏

ganes1410:
@tvararu  I need to thank you for your guidance

PR #47, opened 2017-04-07
Pass through unrecognised keys to input

tvararu:
This allows users to continue typing when they are focusing an option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 _(add items here for easier creation of next log entry)_
 
 - Add `showNoOptionsFound` property to allow users to disable this behaviour.
+- Pass through unrecognised key events to input, allowing users to continue typing when they are focusing an option.
 
 ## 0.3.5 - 2017-04-06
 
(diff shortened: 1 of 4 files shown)


tvararu:
@edwardhorsford could I ask you to confirm if this is implemented correctly from a user's perspective? The input has the yellow border even when it's not the currently selected element, but no caret. Also, any key other than the recognised ones will set focus to the input, including modifier ones like `cmd` or `option`.

edwardhorsford:
Theo points out with the current suggestion, using modifier keys will also move the focus, which may be *weird*. I'd like to try passing through *just* alpha-numeric keys (keycodes), and leaving all else. Would need to check that shift-k still passes through as "K" though...

Feels like it will need some good testing.

tvararu:
@edwardhorsford here's the list of accepted "printable characters" that the typeahead will forward: https://github.com/alphagov/accessible-typeahead/pull/47/files#diff-cd432b59e6c3f8e0579c8aad2b4eec7cR15

I have tested that this works how you would expect with modifier keys and I think it's good. 👍

PR #50, opened 2017-04-08
Write tests for wrapper

nickcolley:
Increased the coverage by testing the Wrapper.jsx entrypoint.

- Tested what's in the README for enhancing select elements
- Setting `opt.defaultValue` now will update the original select elements' selected option
- `opt.defaultValue` will now be set by the original select elements' selected option unless overridden
- Ensure original select elements' name is transfered correctly to the typeahead
- Tested that onSelect updates the original select element

Before:

image

After:

image

### DIFF
--- a/test/browser/wrapper.jsx
+++ b/test/browser/wrapper.jsx
@@ -1,9 +1,45 @@
 /* global describe, expect, it */
 import Wrapper from '../../src/wrapper'
 
+const injectSelectToEnhanceIntoDOM = (id) => {
+  const options = {
+    fr: 'France',
+    de: 'Germany',
+    gb: 'United Kingdom'
+  }
+
+  var $select = document.createElement('select')
+  $select.id = id
+  $select.name = id
+
+  for (var optionKey in options) {
+    var option = document.createElement('option')
+    option.value = optionKey
+    option.text = options[optionKey]
+    $select.appendChild(option)
+  }
+
+  document.body.appendChild($select)
+}
+
 describe('Wrapper', () => {
   it('exposes global on window', () => {
     expect(typeof Wrapper).to.equal('object')
     expect(typeof window.AccessibleTypeahead).to.equal('function')
+  })
+  it('can enhance a select element', () => {
+    const id = 'location-picker'
+    injectSelectToEnhanceIntoDOM(id)
+
+    window.AccessibleTypeahead.enhanceSelectElement({
+      selectElement: document.querySelector('#' + id)
+    })
+
+    const typeaheadInstances = document.querySelectorAll('.typeahead__wrapper')
+    expect(typeaheadInstances.length).to.equal(1)
+
+    const typeaheadInstance = typeaheadInstances[0]
+    expect(typeaheadInstance.innerHTML).to.contain(`id="${id}"`)
+    expect(typeaheadInstance.innerHTML).to.contain(`name="${id}"`)
   })
 })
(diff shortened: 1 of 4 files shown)


nickcolley:
Need to update https://github.com/alphagov/accessible-typeahead/blob/master/README.md#progressive-enhancement too...

test/browser/wrapper.jsx
```diff
@@ -0,0 +1,181 @@
...
+const injectSelectToEnhanceIntoDOM = (element, id, name, options, selectedOption) => {
+  var $select = document.createElement('select')
+  if (id) {
+    $select.id = id
+  }
+  if (name) {
+    $select.name = name
+  }
+
+  for (var optionKey in options) {
```
tvararu: Only if you want™:

You can switch this to `Object.keys(options).map(foo).forEach(bar)` which separates the array transformation logic from the `appendChild` side effect.

test/browser/wrapper.jsx
```diff
@@ -0,0 +1,181 @@
...
+  beforeEach(() => {
+    scratch.innerHTML = ''
+  })
+
+  after(() => {
+    scratch.parentNode.removeChild(scratch)
+    scratch = null
+  })
+
+  it('exposes global on window', () => {
```
tvararu: Nice. 👍

test/browser/wrapper.jsx
```diff
@@ -0,0 +1,181 @@
...
+    window.AccessibleTypeahead.enhanceSelectElement({
+      selectElement: document.querySelector('#' + id)
+    })
+
+    const typeaheadInstances = document.querySelectorAll('.typeahead__wrapper')
+    const typeaheadInstance = typeaheadInstances[0]
+    const typeaheadInput = typeaheadInstance.querySelector('.typeahead__input')
+    const typeaheadOption = typeaheadInstance.querySelector('.typeahead__option')
+
+    // Using setTimeouts here since changes in values take a while to reflect in lists
```
tvararu: Hmm I'm really keen to find a better way to do this, I want to avoid slow tests if possible.

nickcolley: Same, this felt v flakey (with too much knowledge of the internal polling) but wanted to get the PR infront of you asap.

test/browser/wrapper.jsx
nickcolley: I don't think this is worth it, you just end up with duplicated code since DOM manipulations are not functional.

nickcolley:
Reading the docs it seem as if setting the name to '' is intentional?

> - Default the typeahead `name` attribute to `''` to prevent it being included in form submissions

test/browser/wrapper.jsx
tvararu: Not sure what you mean by duplication? I meant:

```js
Object.keys(options)
  .map(optionKey => {
    const option = document.createElement('option')
    option.value = optionKey
    option.text = options[optionKey]
    option.selected = (selectedOption === optionKey)
    return option
  })
  .forEach(option => $select.appendChild(option))
```

Now that I've written it I have a strong preference for this, because I realised it would keep our codebase for-loop-free™.

tvararu:
> Reading the docs it seem as if setting the name to '' is intentional?

It is. The typeahead in that scenario should not be submitted as part of the form; only the side effect that modifies the hidden `<select>` should matter.

tvararu:
Closing in favour of https://github.com/alphagov/accessible-typeahead/pull/108.

PR #52, opened 2017-04-08
Update example page styles so it's easier to read

nickcolley:
There's some content work to do for this page as at the moment it's a bit cryptic but this PR makes the current docs easier to read.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -7,11 +7,11 @@
     <style>
       /* Example page specific styling. */
       html {
+        color: #111;
+        background: #FFF;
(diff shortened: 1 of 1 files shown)


nickcolley:
I prefer to group rather than alphabetical, means similar changed are not across many lines.

Agree with the casing.

nickcolley:
I prefer to group rather than alphabetical, means similar changes are not across many lines.

Agree with the casing.

tvararu:
I prefer alphabetical but I don't really care much about linting unless it's automated.

Tested on Safari / Chrome and looks improved, thanks! 👍

PR #69, opened 2017-04-25
Only render hint input if there is something to hint

### LINKED ISSUE #56: Hint is selectable on some screen readers and browsers
The typeahead [contains a `<Hint>` element](https://github.com/alphagov/accessible-typeahead/blob/d2d2181adb591f6d6707caacf21ae706fe0a0cf0/src/typeahead.jsx#L398) that is responsible for displaying the grey text in this image:

Image of the grey typeahead hint

I've done some accessibility testing and on JAWS17 Chrome, and NVDA Chrome/FireFox/IE11, it is possible to sometimes tab to the hint input element (even though it has `tabindex="-1"`!), in which case the user will be told it's "read-only edit" and not be allowed to edit it. This only seems to occur when the screen readers are running.

There's a few solutions to try but one would be to simply not display the `<Hint>` unless it's in use.

tvararu:
This prevents JAWS and NVDA from picking it up on initial page load as a possible form element, which they do in spite of the `tabindex="-1"`.

Closes #56.

Tested:

- [x] iOS VoiceOver
- [x] IE11 JAWS17
- [x] IE11 NVDA

### DIFF
--- a/src/typeahead.js
+++ b/src/typeahead.js
@@ -355,12 +355,14 @@ export default class Typeahead extends Component {
       const hintValue = (optionBeginsWithQuery && autoselect)
         ? query + selectedOption.substr(query.length)
         : ''
-      return <input
-        className={`${cssNamespace}__hint`}
-        readonly
-        tabindex='-1'
-        value={hintValue}
-      />
+      return hintValue
+        ? <input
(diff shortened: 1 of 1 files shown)


tvararu:
Added a "Do not merge" label until I do further testing with AT.

PR #70, opened 2017-04-25
Move aria live region above input

tvararu:
This allows screen reader users to more easily discover it.

- [x] IE11 NVDA (works as expected)
- [x] iOS VoiceOver (swiping seems to skip the label so it's slightly confusing)

### DIFF
--- a/src/typeahead.js
+++ b/src/typeahead.js
@@ -430,6 +430,12 @@ export default class Typeahead extends Component {
 
     return (
       <Wrapper>
+        <Status
+          length={options.length}
+          queryLength={query.length}
+          minQueryLength={minLength}
+          selectedOption={this.templateInputValue(options[selected])}
+        />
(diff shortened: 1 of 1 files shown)


adamsilver:
> This allows screen reader users to more easily discover it.

How so?

tvararu:
@adamsilver we had a user on IE11 JAWS16 that managed to move focus under the input; what they heard was (roughly):

```
(label) What was the last country you visited?, label
(input) What was..., input, editable
(region) Type in 2 or more characters for results.
```

It seems like this would be more intuitive:

```
(label) What was the last country you visited?
(region) Type in 2 or more characters for results.
(input) What was..., input, editable
```

tvararu:
Added a "Do not merge" label until I do further testing with AT.

adamsilver:
Niiice. Thanks!

PR #80, opened 2017-05-03
Add "style" declaration in package.json

lennym:
At the Home Office we use a sass compiler that reads from a "style" declaration in package.json files of npm installed modules when compiling css.

By adding a "style" declaration we can include this module's css into our built and compiled code more easily.

Before:

```
@import "accessible-typeahead/dist/styled.min.css"
```

After:

```
@import "accessible-typeahead"
```

It's a pretty minor thing, but smoothes the developer experience as no knowledge of the package internal structure is required to implement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 _(add items here for easier creation of next log entry)_
 
 - Add touchEnd handler for iOS and touch devices, fixes custom suggestions.
+- Add `style` declaration in package.json
 
 ## 0.4.1 - 2017-04-26
 
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "accessible-typeahead",
   "version": "0.4.1",
   "main": "dist/accessible-typeahead.min.js",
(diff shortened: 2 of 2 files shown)


tvararu:
Didn't know you could do this, great!

If you could do the following (like in your other PR) I'll ship this ASAP:

- [ ] Update README
- [ ] Update CHANGELOG

(I'll update the CONTRIBUTING to document this when I get a moment! 🌈 )

lennym:
Well, it's very dependent on the exact sass compiler being used. In the case of HO projects, we use [npm-sass](http://npmjs.com/npm-sass), which is an extension of node-sass with npm module resolution.

I don't know what other compilers will make use of the declaration, but it's certainly a pattern that exists elsewhere - e.g. https://github.com/twbs/bootstrap/blob/v4-dev/package.json#L61-L62

tvararu:
@lennym hmm, then we should be fine without a README explanation because as you point out there's many ways to do it.

lennym:
Now with added changelog!

PR #81, opened 2017-05-03
Use UMD module definition

lennym:
This will allow the module to be used both standalone in browser (as per the README documentation) but also compiled into an application bundle when installed into a project from npm (i.e. using browserify or webpack in a parent project).

Note the `module.exports` in wrapper.jsx - it would be probably preferable for this to be `export default`, but this results in the module being wrapped in a `{ __esmodule: true, default: ... }` construct by webpack, which makes a mess of the API. By using `module.exports` instead then the exported module is a direct reference to `AccessibleTypeahead`.


tvararu:
Thanks for this @lennym!

Not exposing a clean `module.exports` interface was an intentional oversight during initial development, as the pattern was still experimental.

Your timing is impeccable as we've only just now started work on finalising a `v1.0.0`. I was going to open a milestone tomorrow and create issues for the steps leading up to that, which includes this work, that you've already done for me. :clap:

Can you update this PR with:

- [ ] A README section tutorialising how to use this? (`npm install accessible-typeahead && etc`)
- [ ] A CHANGELOG bullet detailing the change being made.

Do note: the `accessible-typeahead.min.js` file [bundles up dependencies](https://github.com/alphagov/accessible-typeahead/blob/9d66de7b413294f2dcb50a9cfa2c26ae2c05ccb0/package.json#L31-L34). If you're using this as part of your own build pipeline that can lead to including libraries twice. Happy to tackle this separately but we do want to offer a solution for this as part of v1.0.0.

For v1.0.0 I'd also I'd like to support using this in a project that already has `(p)react` seamlessly. In short all of these should exist (work to be done separately from this PR):

```html
<!-- Bundles dependencies. -->
<script src="accessible-typeahead.bundle.min.js"></script>
<script>
  AccessibleTypeahead()
</script>

<!-- Standalone build without all the dependencies. -->
<script src="preact.js"></script>
<script src="lodash.debounce.js"></script>
<script src="accessible-typeahead.standalone.min.js"></script>
<script>
  AccessibleTypeahead()
</script>
```

```js
import AccessibleTypeahead from 'accessible-typeahead'
AccessibleTypeahead()

// ---

import {h} from 'preact' /** @jsx h */
import AccessibleTypeahead from 'accessible-typeahead/preact'

<AccessibleTypeahead />

// ---

import React from 'react'
import AccessibleTypeahead from 'accessible-typeahead/react'

<AccessibleTypeahead />
```

lennym:
Yes, I did think about the issue of potentially duplicated bundled dependencies when I was writing, but figured given the warning in the top of the README that "working at all" was a better bet than trying to get fully optimal on the first pass.

I suspect that the best solution in the longer term is likely to be to compile slightly different versions for different install targets. In the immediate term I think possibly there might be value in having two (or possibly three) output files:

* One for direct `<script>` embedding in the browser - may or may not have bundled dependencies
* One for compiling using browserify/webpack - would not need bundled dependencies since the compiler can resolve the transitive dependencies, but should not require any babel/jsx configuration in the end project. 

The latter is the use case I'm primarily interested in supporting for our own ends. In particular I'm keen to be able to `const Typeahead = require('accessible-typeahead')` in our HO projects.

lennym:
I've added a note to the README and CHANGELOG covering the work done so far.

I'm happy to try to help out with optimising these built/release toolchains a bit, but with a huge disclaimer that I'm primarily a node dev, and am such a bit behind the curve when it comes to babel/jsx/react/webpack compilation (my usual approach is to sidestep the issue by never writing "modern" browser code).

tvararu:
Rebased this in https://github.com/alphagov/accessible-typeahead/pull/86, going to close this if that's ok!

PR #85, opened 2017-05-03
Add touchEnd handler for iOS and touch devices

tvararu:
Custom suggestions don't seem to bubble events correctly if you just use the native browser `click` handler on the parent. `touchEnd` does seem to work.

This allows iOS to reliably click on custom suggestions.

### DIFF
--- a/src/typeahead.jsx
+++ b/src/typeahead.jsx
@@ -9,7 +9,7 @@ const kc = {
 }
 
 function isIosDevice () {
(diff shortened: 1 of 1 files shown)


nickcolley:
Should we revert this?

tvararu:
@nickcolley Not sure, this introduced a new bug while fixing another. Arguably not being able to scroll on mobile is worse than not being able to correctly use the "custom suggestions" feature which is very small by comparison. So I'm also leaning on reverting.

PR #86, opened 2017-05-03
Use UMD module definition

tvararu:
Just rebasing https://github.com/alphagov/accessible-typeahead/pull/81.


tvararu:
@lennym I've published this as `v0.4.2`!

Re: helping out we'll happily take further contributions. 👏

PR #88, opened 2017-05-03
Update readme to discourage use of `placeholder`

### LINKED ISSUE #41: Don't provide a placeholder option
We shouldn't encourage the use of the placeholder attribute. Here's some details:

[https://govuk-elements.herokuapp.com/form-elements/#form-hint-text](https://govuk-elements.herokuapp.com/form-elements/#form-hint-text)

And some more details as to why:

[http://adamsilver.io/articles/placeholders-are-problematic/](http://adamsilver.io/articles/placeholders-are-problematic/)

tvararu:
Closes #41.

@adamsilver what do you think of this?


adamsilver:
Yep, looks good.  And thanks for including the link, @tvararu.

PR #108, opened 2017-05-09
Write tests for wrapper

tvararu:
Previously https://github.com/alphagov/accessible-typeahead/pull/50.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 _(add items here for easier creation of next log entry)_
 
+- Default `defaultValue` when progressively enhancing.
+
 ## 0.5.0 - 2017-05-09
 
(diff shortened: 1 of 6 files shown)


tvararu:
@nickcolley I decided to leave the `setTimeout` hack in because in the future I'd like to write our tests with Jest so we can use [timer mocks](https://facebook.github.io/jest/docs/en/timer-mocks.html#content).

nickcolley:
Are you sure we want these tests if we have your new e2e tests?

tvararu:
@nickcolley Yes, they are faster. The end to end tests are mainly responsible for keeping a finger on multi-browser support.

tvararu:
Thanks again for your contribution @nickcolley. 👍

nickcolley:
image

PR #111, opened 2017-05-10
Better errors

tvararu:
Throw more errors when required arguments are not passed into `accessibleTypeahead` or `enhanceSelectElement`.

Slightly update the README based on feedback from @nickcolley.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,14 @@ Or using a script tag:
 And then call the `accessibleTypeahead` function, providing a suggestion engine:
 
 ```js
-function suggest (query, syncResults) {
+function suggest (query, populateResults) {
   const results = [
     'France',
     'Germany',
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -43,13 +43,13 @@ Or using a script tag:
...
 ```js
-function suggest (query, syncResults) {
+function suggest (query, populateResults) {
   const results = [
     'France',
     'Germany',
     'United Kingdom'
   ]
-  syncResults(results.filter(result => result.indexOf(query) !== -1))
+  populateResults(results.filter(result => result.indexOf(query) !== -1))
```
nickcolley: could split this into two lines?

```
filteredResults = [...]
populateResults(filteredResults)
```

tvararu: Done

Issue #115, opened 2017-05-11
Provide a built-in suggestion engine


tvararu:
The [progressive enhancement](https://github.com/alphagov/accessible-autocomplete#progressive-enhancement) approach already uses a very simple suggestion engine. It would be easier to use the autocomplete if it offered an API that was as simple as:

```js
accessibleAutocomplete({
  element: document.querySelector('#autocomplete-container'),
  options: ['United Kingdom', 'United States']
})
```

And as a later feature, supporting (at a basic level):

- custom weighting
- synonyms / aliases (e.g. UK for United Kingdom)

Some work on this can be based on the [openregister-picker-engine](https://github.com/alphagov/openregister-picker-engine).

tvararu:
I have updated the title of this issue now that https://github.com/alphagov/accessible-autocomplete/pull/154 is merged. The more advanced features are still up for grabs (we may split this into multiple issues later).

edwardhorsford:
It might be good to consider an API similar to the [baymard picker](https://baymard.com/labs/country-selector) that allows you to provide alternate spellings as part of the select using data attributes.

alexnewmannn:
this is definitely something we are keen to have in DWP, we require a list of hospitals and their aliases both of which should return results. We may also require address to return results, will be happy to pick it up if we have time within our team 👍

PR #127, opened 2017-05-12
Update `style` property in package.json to reflect updated filename

lennym:
Trying to compile some sass which includes this module fails because the referenced file in package.json no longer exists. Update the reference to point to the new file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "accessible-autocomplete",
   "version": "1.0.0",
   "main": "dist/accessible-autocomplete.min.js",
-  "style": "dist/styled.min.css",
+  "style": "dist/accessible-autocomplete.min.css",
   "description": "An autocomplete component, built to be accessible.",
   "repository": "alphagov/accessible-autocomplete",
   "author": "Theodor Vararu <theodor.vararu@digital.cabinet-office.gov.uk>",


tvararu:
Thanks @lennym, I forgot to update this. I'll ship it ASAP. 👍

tvararu:
Published as `v1.0.1`!

lennym:
No problem, thanks for getting on it quickly.

I'm working on a spike to integrate this in place of our existing autocomplete functionality and I'm hitting a couple of snags so you might get a few more PRs from me this afternoon.

FYI - https://github.com/UKHomeOffice/firearms/compare/spike/gds-autocomplete

tvararu:
@lennym no problem, happy to help!

Please submit issues or fixes with anything you encounter. 👍 

I see your existing implementation uses `Bloodhound` which is much smarter than [the basic engine we use when progressively enhancing](https://github.com/alphagov/accessible-autocomplete/blob/0d70f85659e69a694522dcc1303b0e7a3174c44d/src/wrapper.js#L12-L17). We have [plans to provide a better built-in engine](https://github.com/alphagov/accessible-autocomplete/issues/115), but until that point, you have two options:

- Continue using Bloodhound: `source: (new Bloodhound(opts)).search`
- Modify our `createSimpleEngine` with additional sorting/filtering based on your needs

PR #131, opened 2017-05-15
Do not copy select element name attribute to autocomplete input

### LINKED ISSUE #130: `enhanceSelectElement` does not unset `name` as claimed in README
The README section for `enhanceSelectElement` states the following:

> Default the autocomplete name attribute to '' to prevent it being included in form submissions

However the code at https://github.com/alphagov/accessible-autocomplete/blob/master/src/wrapper.js#L36 sets it to the same name as the select element unless explicitly overwritten - and it cannot be explicitly set to `''` since that is a falsy value (as per #128)

I'm not sure whether the docs or the code are correct in this instance (though my instinct is the code) so haven't PR'ed a fix.

lennym:
Copying the name attribute means that the same `name` fields is POSTed twice on form submission. This can cause server-side parsers to fail because they receive two fields with the same key. By not copying the name attribute onto the autocomplete element we can ensure that only one value is POSTed to the server on form submission.

Fixes #130


tvararu:
Published in `v1.0.3`!

PR #132, opened 2017-05-15
Make integration tests run on Travis !SAUCE_ENABLED

tvararu:
This should allow us to do integration testing on PRs from forks.

In addition to this, I've changed the `SAUCE_ENABLED` variable to be a private one on Travis to prevent it from being set to true on fork PRs.

See:

- https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI
- dwyl/learn-nightwatch#8

PR #133, opened 2017-05-15
Add a null option support for `enhanceSelectElement`

lennym:
Allows the common pattern for `<select>` elements in forms to have a default option with no value and a placeholder text to be supported out of the box.

### Changes
* Uses `defaultValue` only when the `<select>` element has no selected value (i.e. when a null option is selected)
* Filters out any null-valued options from the option list used to populate the `source` array for autocomplete
* Adds a flag to optionally not apply the above filter

### Tests
* Refactors the `enhanceSelectElement` tests to reduce repetition of setup code
* Removes some of the test scope for `can enhance a select element` on the basis that it's too vague to be helpful. Assertions have been split out into more specific tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 _(add items here for easier creation of next log entry)_
+- Add support for handling null/placeholder options when using `enhanceSelectElement`
 
 ## 1.0.3 - 2017-05-15
 
--- a/README.md
+++ b/README.md
@@ -224,6 +224,32 @@ This function takes the same options as `accessibleAutocomplete`, with the only 
 
 > **Note**: The `accessibleAutocomplete.enhanceSelectElement` function is fairly light and wraps the public API for `accessibleAutocomplete`. If your use case doesn't fit the above defaults, try [reading the source](src/wrapper.js) and seeing if you can write your own.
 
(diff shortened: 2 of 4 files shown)


tvararu:
@lennym can you please rebase this now that I've merged https://github.com/alphagov/accessible-autocomplete/pull/132? Hopefully that should green this build!

lennym:
Rebased and pushed. If you're broadly happy with my changeset I'll update the CHANGELOG for you as well.

tvararu:
@lennym go ahead!

lennym:
Done. 

I'm not sure what your feelings are on whether it's a breaking change or not. To my mind it's a fix of broken functionality around `defaultValue` in that context, so would not be a breaking change, but that's coming from a position of privilege in that I don't have any code in prod, and I 100% understand that people relying on the previous behaviour might well disagree.

tvararu:
@lennym I agree with your line of thinking that it's bringing the functionality of the `defaultValue` option closer to what is expected out of the box. It's a breaking change if you rely on the previous behaviour, which is equivalent to setting `preserveNullOptions: true`.

Happy to ship this as a patch release and take full responsibility if this *does* break something for anyone.

tvararu:
Published in `v1.0.4`!

PR #136, opened 2017-05-15
Add a pre commit compile hook

### LINKED ISSUE #135: v1.0.4 not built before publish
I think this one is 100% my bad, but I think the compiled code in v1.0.4 doesn't match the build output from the source. It's possible/probably that I forgot to build or forgot to add the built code before raising my PR. Running the npm version gives different results to when I have a local version `npm link`ed into my project.

Normally I'd suggest protecting against this by compiling in pre-publish hook for npm, but that doesn't solve the fact that some use cases will want the compiled code in github. Given that, I'm working on a pre-commit hook that will keep things synced.

Meanwhile, you might want to rebuild and publish a new version.

Sorry!

lennym:
Prevent doing things like I did with #135 by compiling the code pre-commit and then checking that the diff is empty before allowing the commit to continue.

This ensures that the compiled version always matches the source. Hopefully.

Fixes #135


package.json
```diff
@@ -28,7 +28,8 @@
     "standard": "standard",
     "test": "run-p standard karma wdio",
     "version": "npm build && git add -A dist",
```
tvararu: ^ But we already have this? o_O

tvararu:
This is very strange, we already have the `npm version` script that will always compile and commit a latest dist. It's worked so far, so I don't think this should be necessary. (though I would like to get CI to do it instead of me)

I'll look tomorrow into it to see what's happened with 1.0.4 and if necessary we can deprecate it and publish a new one. Don't sweat it. 👌

tvararu:
I scrolled my terminal output and basically I'm not sure what happened to the `npm build` script. There's no output:

screen shot 2017-05-15 at 18 17 57

I agree we need a few additional checks. 😉 

Either way, this is definitely not your fault, but likely some gap in our scripts (which I wrote, so assign blame to me). We can fix them.

lennym:
The short answer is that `npm build` is not equivalent to `npm run build`. The former doesn't run the `build` script from your scripts.

I'd probably be in favour of the pre-commit hook anyway, since then the `dist` directory and the source code are "guaranteed" to be in sync on any commit, and not just versions.

package.json
lennym: See my comment below. This needs to be `npm run build`.

tvararu:
I'm thinking about this but I'm not sure I want to build on pre-commit. While it's not something I stick to religiously, for longer PRs I like to do lots of small commits with just one `Update dist` commit at the end, so reviewing is simpler. You can `--no-verify` but then it becomes too tempting to use that at all times, which defeats the purpose.

We definitely need something in place to make sure that code that reaches master always has the latest compiled dist, so that our examples pages don't break.

How about on `prepush`?

I think we should use [husky](https://github.com/typicode/husky) instead of `pre-commit` for this. (doesn't require package.json config)

tvararu:
Closing in favour of #148.

PR #137, opened 2017-05-15
Fix `version` script to compile assets

lennym:
`npm build` is not equivalent to `npm run build` - we need the latter here.


tvararu:
Learned something new. 👍

PR #140, opened 2017-05-16
Support importing the library as a standard React/Preact module

### LINKED ISSUE #89: Support more ways of using the library
Here are a few:

## Unbundled build

```html
<script src="preact.js"></script>
<script src="lodash.debounce.js"></script>
<!-- Standalone build without all the dependencies. -->
<script src="accessible-typeahead.standalone.min.js"></script>
<script>
  AccessibleTypeahead()
</script>
```

## JS module without all the bundled dependencies

```js
// In package.json
{
  "dependencies": {
    "lodash.debounce": "*",
    "preact": "*",
    // ...
  }
}
// Then maybe?
import AccessibleTypeahead from 'accessible-typeahead/standalone'

AccessibleTypeahead()
```

## Preact module

```js
import {h} from 'preact' /** @jsx h */
import AccessibleTypeahead from 'accessible-typeahead/preact'

<AccessibleTypeahead />
```

## React module

```js
import React from 'react'
import AccessibleTypeahead from 'accessible-typeahead/react'

<AccessibleTypeahead />
```

tvararu:
Closes #89.


test/functional/index.js
```diff
@@ -1,5 +1,5 @@
 /* global after, describe, before, beforeEach, expect, it */
-import { h, render } from 'preact' /** @jsx h */
+import { createElement, render } from 'preact' /** @jsx createElement */
```
nickcolley: I think we can get rid of these JSX comments?

tvararu: If we do, the compilers will still work (possibly not for the React bundle), but Standard.js will complain that it's an unused variable.

tvararu:
> Would we then publish standalone, react, preact to npm? This is looking cool.

All 3 as part of the same package.

> What's the use case for us building the preact/react?

So by importing `autocomplete.js` directly from inside our package? Users can do that but because their transpilation steps may vary (most exclude `/node_modules/`), their Node versions may vary, they may be including them straight into the browser, and so on. It's better to compile using babel and provide a dist. Lots of libraries do this:

- https://unpkg.com/react-motion@0.5.0/lib/Motion.js
- https://unpkg.com/next@2.3.1/dist/lib/head

Issue #59, opened 2017-04-10
Add support for "required" attribute


tvararu:
Make it possible to pass an argument to the `AccessibleTypeahead` function to set the [`required` attribute](https://www.wufoo.com/html5/attributes/09-required.html) on the typeahead input field. This is useful if the need involves free text submission.

samtsai:
@tvararu Good to close this?

tvararu:
@samtsai yes! Forgot. 🌈

PR #142, opened 2017-05-17
Add required option

samtsai:
Addresses issue #59


test/functional/index.js
```diff
@@ -44,6 +44,18 @@ describe('Autocomplete', () => {
...
       })
 
+      it('renders an input with a required attribute', () => {
+        render(<Autocomplete required />, scratch)
+
+        expect(scratch.innerHTML).to.contain('required')
+      })
+
+      it('renders an input without a required attribute', () => {
+        render(<Autocomplete required={false} />, scratch)
```
nickcolley: Should we also have a test that checks the required state if no prop is set at all?

tvararu: I think that would be testing the vdom library and not our own code, so I don't think it's necessary.

tvararu:
Published in `v1.1.0`!

PR #147, opened 2017-05-24
fix progressive enhancement error in firefox < v48

revilossor:
The progressive enhancement doesnt work in firefox 45 - this PR replaces an [Element.insertAdjacentElement](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement) call in wrapper.js (which is only implemented in firefox 48+) with [Node.insertBefore](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) which will work in older versions of firefox.


tvararu:
Cool! 👍 

Not sure we need the `.nvmrc` but it does no harm.

tvararu:
Published in `v1.2.1`!

PR #154, opened 2017-06-01
Use a default search engine

joelanman:
Previously, a `source` function was required.

This PR simplifies so that only an array of `values` is required. In this case, a simple search engine is used to search the array for whatever's been typed.

Optionally a `source` function can be provided as before.


src/autocomplete.js
```diff
@@ -18,6 +18,13 @@ const hasPointerEvents = (() => {
   return element.style.pointerEvents === 'auto'
 })()
 
+const createSimpleEngine = results => (query, syncResults) => {
```
tvararu: Later (not now) we will want to put this into its own module and write some tests for it.

joelanman:
addresses #115

src/autocomplete.js
```diff
@@ -51,14 +58,19 @@ export default class Autocomplete extends Component {
...
+    values: []
   }
 
   elementRefs = {}
 
   constructor (props) {
     super(props)
 
+    if (!props.source) {
+      props.source = createSimpleEngine(props.values)
```
tvararu: Interesting. Reassigning `props` is not recommended, however this is setting a default in a way that isn't possible using `defaultProps`.

I think this is fine.

src/autocomplete.js
```diff
@@ -51,14 +58,19 @@ export default class Autocomplete extends Component {
...
+    required: false,
+    values: []
   }
 
   elementRefs = {}
 
   constructor (props) {
     super(props)
 
+    if (!props.source) {
```
tvararu: Maybe this is clearer:

```js
const setDefaultSource = !props.source && props.values.length
if (setDefaultSource)
```

src/wrapper.js
```diff
@@ -4,26 +4,18 @@ import Autocomplete from './autocomplete'
...
-    : []
-  syncResults(filteredResults)
-}
-
 accessibleAutocomplete.enhanceSelectElement = (opts) => {
   if (!opts.selectElement) { throw new Error('selectElement is not defined') }
 
   // Set defaults.
-  if (!opts.source) {
+  if (!opts.values) {
```
tvararu: This needs a README change:

```
- Default the autocomplete `source` to use existing `<option>`s from the `<select>`
```

To:

```
- Default the autocomplete `values` to use existing `<option>`s from the `<select>`
```

src/wrapper.js
```diff
@@ -4,26 +4,18 @@ import Autocomplete from './autocomplete'
 function accessibleAutocomplete (opts) {
   if (!opts.element) { throw new Error('element is not defined') }
   if (!opts.id) { throw new Error('id is not defined') }
-  if (!opts.source) { throw new Error('source is not defined') }
+  if (!opts.source && !opts.values) { throw new Error('values or source are not defined') }
```
tvararu: This needs a README change; add a header and short description for this new [Required option](https://github.com/alphagov/accessible-autocomplete#required-options).

Also update the `source` description so it's clear that you can specify one or the other.

tvararu: Also, come to think of it, update the [usage examples](https://github.com/alphagov/accessible-autocomplete#using-npm-and-a-module-system) to use this new syntax, because it's much friendlier.

tvararu:
Thinking: if you have both `source` and `values` specified, then the latter will be ignored. This isn't great.

What do you think about overloading the `source` option for this feature? So these are the same:

```js
accessibleAutocomplete({
  element: el,
  source: mySourceFoo
})

accessibleAutocomplete({
  element: el,
  source: ["United Kingdom", "United States"]
})
```

src/autocomplete.js
tvararu: Scratch that (sorry!), this isn't correct. Instead, do this in `wrapper.js`:

```diff
 function accessibleAutocomplete (opts) {
    if (!opts.element) { throw new Error('element is not defined') }
    if (!opts.id) { throw new Error('id is not defined') }
-   if (!opts.source) { throw new Error('source is not defined') }
+   if (!opts.source) {
+     if (!opts.values) {  throw new Error('values or source are not not defined') }
+     opts.source = createSimpleEngine(opts.values)
+   }
  
    render(<Autocomplete {...opts} />, opts.element)
  }

nickcolley:
Sweet! I think we need to add additional tests for this default behaviour too

joelanman:
cool yeh overloading sounds good, I'll make the changes

joelanman:
hm somethings breaking my brain a bit so writing it here. In the case of progressively enhancing a `<select>`, we'd probably want to support a custom search engine, but use the values in the `select`. Can a custom function passed to `source` access the values in the `select`?

tvararu:
@joelanman Once you specify a custom engine it's up to you to get the values. For that use case, end users can query `selectElement.options` like we do.

Specifying a `source` function when using `enhanceSelectElement` isn't a way to inject a better version of `createSimpleEngine` while using the same `values` we inferred. Rather, it's a way to override everything we inferred with a whole new `createSimpleEngine(values)`.

`enhanceSelectElement` also comes with the caveat that if it doesn't do something exactly to your taste, it's sufficiently short that you can roll your own.

joelanman:
I've updated this to overload `source` - happy to squash the commits so we just get this version if thats what we want.

joelanman:
I've rebased and updated README, not sure about CHANGELOG - do you mind doing that?

tvararu:
@joelanman sure, just one more nit, if you can rebase again to clean up the commit messages (first one still references `values` option which would be confusing later); 3 commits should be fine (the feature, dist update, README)

joelanman:
@tvararu  OK I think it's all good now!

tvararu:
Published in `v1.3.0`!

Issue #31, opened 2017-03-28
Functionality: Down key show all results?


nickcolley:
Not sure if this is a good idea, but when I use keyboard input on a select input, then press down on the keyboard, it'll open the select and let me go through them.

tvararu:
I don't see why we shouldn't do this, @edwardhorsford what do you think?

nickcolley:
This could cause performance issues if the list is very long though...

tvararu:
@nickcolley there's situations in which you can't do it like if you rely on async results (which are not built yet), or indeed if the total list of results is too large. If the list is something like the days of the week, "Monday" to "Sunday", I think it could be good to have a way to bring up the list even before the user types anything.

edwardhorsford:
I have no immediate objection to the idea, but with the current styling don't expect any users to actually do this.

If we try it I'd want to make sure selection works as expected.

nickcolley:
> If we try it I'd want to make sure selection works as expected.

What do you mean by 'selection'?

adamsilver:
I think we should be able to configure the typeahead to show the options within it, like a native select. To make that clear to the user, I think we should have a down arrow (or similar). Clicking it, as well as pressing the DOWN key, should those options.

tvararu:
Right, this is a good feature, and we should provide an option.

Off the top of my head these two things need to be updated to enable this:

- The typeahead would need the ability to call the `source` function even when the query is empty, which is currently a [hardcoded constraint](https://github.com/alphagov/accessible-typeahead/blob/784d74fae7473e0c2eed8e02f27ea6ef21750fa6/src/typeahead.jsx#L177)
- The [down arrow handler](https://github.com/alphagov/accessible-typeahead/blob/784d74fae7473e0c2eed8e02f27ea6ef21750fa6/src/typeahead.jsx#L265) needs to open the menu if it's collapsed and there are available options

The tests will need to be updated to reflect the new behaviours too, and it will also need an example (like picking from Monday to Sunday).

I'm currently prioritising productisation features, but if anyone else wants to pick it up I'm happy to provide further guidance.

edwardhorsford:
@adamsilver Could you describe an example use case?

I think it's something that may be useful, but would be good to work on the design and test in user research - I have the feeling that it could introduce some usability issues.

adamsilver:
Hi @edwardhorsford . The behaviour I am talking about is basicaly the HTML5 datalist but not a buggy version.

[...]

PR #158, opened 2017-06-02
DO NOT MERGE Add option to show all values on click (like a normal dropdown)

joelanman:
for discussion - this PR adds an option `showAllValues`. If this is set to `true`, all values are shown when a user clicks the input, like a normal dropdown. This allows the user to browse the available options.

I think this option might need a visual change to be more accessible - when clicked the list shows, but with keyboard use, you have to press space to see the list, which is non-obvious.


edwardhorsford:
Have given it a quick play. Interesting!

I could see this being useful particularly for shorter lists.

Some quick thoughts:
* The name of the option doesn't seem clear. "DisplayMenuOnFocus" perhaps? I appreciate it's not always done on focus though.
* Would be good to do some research to see if people will still search when this comes up, or whether they think they have to use the menu. As @joelanman says, a different design may help.
* It doesn't play nicely with `autoselect: true`.
* Should down arrow also open the menu?
* Pressing space to see the list actually types a space character

joelanman:
the problem with `autoselect` is that it selects the first item when the input is clicked. Suggest: we could change `autoselect` to only trigger on at least one character.

joelanman:
todo: support up and down keys triggering the open menu

edwardhorsford:
Another possible solution to autoselect:

Do autoselect the first item, but don't display it in the input. If blur or esc, then don't pick the item.

edwardhorsford:
Might want to think about whether we need to tell screen reader users anything different when using this.

src/autocomplete.js
```diff
@@ -206,7 +207,7 @@ export default class Autocomplete extends Component {
 
     this.setState({ query })
 
-    const searchForOptions = !queryEmpty && queryChanged && queryLongEnough
+    const searchForOptions = showAllValues || (!queryEmpty && queryChanged && queryLongEnough)
```
tvararu: I wonder if it makes sense to fold this behaviour into just the `minLength` option? Current `minLength: 0` default does nothing compared to `minLength: 1`, it could mean "show all values" instead.

src/autocomplete.js
```diff
@@ -224,6 +225,10 @@ export default class Autocomplete extends Component {
     }
   }
 
+  handleInputClick (evt) {
+    this.handleInputChange(evt)
```
tvararu: To solve the fact that this doesn't work with keyboard use, what about triggering this in `handleInputFocus`?

src/wrapper.js
```diff
@@ -6,16 +6,19 @@ function accessibleAutocomplete (opts) {
...
 
-const createSimpleEngine = results => (query, syncResults) => {
-  const filteredResults = query
-    ? results.filter(r => r.toLowerCase().indexOf(query.toLowerCase()) !== -1)
-    : []
-  syncResults(filteredResults)
+const createSimpleEngine = (values, showAllValues) => (query, syncResults) => {
+  var matches = []
+
+  if (showAllValues || query !== '') {
```
tvararu: I'm not sure if we even need this (nor did we need it before)! It can just always return something, because the autocomplete won't call it unless it wants something.

Can we just delete the conditional?

tvararu:
> Might want to think about whether we need to tell screen reader users anything different when using this.

@edwardhorsford it probably already works (untested!), if we fold this into `minLength` then I'm pretty confident no further code changes should be required to work with AT.

> todo: support up and down keys triggering the open menu

@joelanman https://github.com/alphagov/accessible-autocomplete/issues/31

src/autocomplete.js
joelanman: Yeah I can see the logic however I think it's not intuitive to just have minlength 0 trigger quite different behaviour to the other min lengths. Especially if we were to look into styling it slightly differently which I think is needed

nickcolley:
Would this make https://github.com/alphagov/accessible-autocomplete/issues/31 easier to implement?

edwardhorsford:
They're very similar.

Like Joe I think we may need a different design. We'll also need to improve how it works with auto-select.

src/autocomplete.js
tvararu: > Yeah I can see the logic however I think it's not intuitive to just have minlength 0 trigger quite different behaviour to the other min lengths.

I don't think the behaviour is different? But I agree that it is obfuscated.

I'm wary of providing conflicting options: `minLength: 3, showAllValues: true` doesn't make sense, because it shows all the options on first click and then hides them again until you get to 3 characters. Some autocompletes do implement a [alwaysRenderSuggestions](https://github.com/moroshko/react-autosuggest#alwaysRenderSuggestionsProp) option but they don't have something like our `minLength`.

I think it makes sense to merge behaviour here under the same option, but `minLength: 0` is weird.

How about replacing `minLength` with this: 

> #### `showValuesMinLength` (default: `1`)
> 
> Type: `number`
> 
> The minimum number of characters that should be entered before the autocomplete will attempt to suggest options. **If this is set to 0, the autocomplete will display suggestions when users first click or tab into it**. When the query length is under this, the aria status region will also provide helpful text to the user informing them they should type in more.

(added ** for emphasis)

We can alias and deprecate `minLength` to this then.

src/autocomplete.js
joelanman: I'm trying to go by how normal dropdowns work, and they don't open on focus. I think if you're tabbing through a page, having stuff open and close is probably not what you want.

joelanman:
I've done two things:

1) remove the check for `query` in `createSimpleEngine` as requested by @tvararu 
2) added a down arrow icon to give keyboard users some affordance that this is a list of options, and if they press space/down/up (last two not implemented yet) they will see those options.

image

completely unsure if I've got the webpack options right for copying the PNG about, and maybe we want it to be SVG

joelanman:
I've put my branch on heroku: https://joelanman-autocomplete.herokuapp.com/examples/

It'd be cool if this repo supported heroku PR review apps, like GOV.UK Elements does:

https://devcenter.heroku.com/articles/github-integration-review-apps

tvararu:
> It'd be cool if this repo supported heroku PR review apps, like GOV.UK Elements does:

It does! But they don't get deployed from Fork apps. You have access to the repo now to do a branch push if you want that though.

src/autocomplete.js
tvararu: Good point!

src/autocomplete.css
```diff
@@ -29,6 +29,14 @@
   outline: 3px solid #ffbf47;
 }
 
+.autocomplete__input.autocomplete__input--show-all-values {
```
tvararu: I don't think we don't need the extra specificity here of the `.autocomplete__input` selector?

webpack.config.babel.js
```diff
@@ -37,12 +37,17 @@ const productionPlugins = [
     // this is actually the property name https://github.com/kimhou/replace-bundle-webpack-plugin/issues/1
     partten: /throw\s+(new\s+)?[a-zA-Z]+Error\s*\(/g,
     replacement: () => 'return;('
-  }])
+  }]),
+
+  new CopyWebpackPlugin([
+    { from: './dropdown-arrow-down.png', to: 'dropdown-arrow-down.png' }
```
tvararu: Your implementation is correct re: copying this to the dist folder for dev testing, and Webpack is notoriously opaque to figure out so kudos!

That being said the issue with providing a separate PNG file is that it's one more thing for users to import and serve in their own application. It's quite a bit of friction.

Two solutions that I feel are better:
- Creating a `src/arrow.js` component that simply returns a very concise blob of inline `<svg>`. The autocomplete is IE9+ only so this is viable;
- In the CSS specify the `background-image` as an inline base64 png.

I think the `<svg>` solution is better because:
- It's much easier to style, you can use `currentColor` to allow users to trivially change the colour for example;
- It's probably smaller, one `<polygon>` element should be enough;
- It scales beautifully regardless of resolutions.

What do you think?

edwardhorsford:
Hi @joelanman - thanks for this! I think it's much better with the arrow.

Some quick thoughts
* Would be good to try with it styled as a button, not an input
* Should clicking the arrow collapse the menu?
* It now works similar to a `select` when first using, but if you pick a value and subsequently return, it works rather differently (not showing the list of options). I wonder if this is an issue.

I'd really like to do some user research with this. I have a concern that we may see that most users forgo the 'autocomplete' bit and instead scroll the list - losing many of the benefits of the component. Doing some research would help us with guidance for when this feature is appropriate to use.

joelanman:
so I think what this PR is aiming for is something like Chosen: https://harvesthq.github.io/chosen/

I agree a button might be the best design in order to fit in with mental models of dropdowns, and yes in that case a second click should close the options.

Good point about filtering the list - again moving to a dropdown style would help. The selected value would appear in the dropdown button part, and the actual text field could be blank - see Chosen for this behaviour.

This is a design pattern which specifically supports the use of browsing through the list. i.e. a situation where a user may well not know what values are available. As such, I'm not really worried about people scrolling instead of typing - that's the intended use case. The typing is to help them search or jump to values if they are aware of them.

Given the changes in behaviour and styling, it may make sense at some point for this PR to actually turn into a fork, or a project that piggy-backs on this project. Or not.

src/autocomplete.css
joelanman: I need it for the padding :) otherwise it's overridden by `.autocomplete__input` later on

edwardhorsford:
@joelanman I'm not worried too, provided it's a suitable use case. My worry is more that we provide appropriate guidance / sensible defaults. A service may decide to use this option because it returns results quicker, not realising how this changes behaviour.

It may be that this component has a section in the guidance for suggested settings depending on use case.

Curious if this would need to be a fork - are they so different? perhaps?

webpack.config.babel.js
joelanman: SVG is probably the answer - just a shame I can't use it as background image and have to pollute the markup with something which isn't that useful for screenreaders

edwardhorsford: @joelanman can you use `aria-hidden="true"` to avoid that issue?

src/autocomplete.css
tvararu: I personally never rely on the cascade or specificity when I use BEM.

I would add another modifier class, like `--default`, which adds the implicit padding and make the code never use both `--default` or `--show-all-values` simultaneously.

So in BEM, I never do this:

```css
.button {
  background-color: blue;
  color: white;
}

.button--primary {
  background-color: green;
}
```

Rather, I do this:

```css
.button {
  color: white;
}

.button--default {
  background-color: blue;
}

.button--primary {
  background-color: green;
}
```

`.button` on its own makes no sense in this example, and it's fine when you have a component system because you can encapsulate this contract neatly away. (rather than relying on users adding the `class="button button--default"` to a million places in their codebase, which is brittle)

webpack.config.babel.js
tvararu: Polluting the markup is not an issue I've encountered before. I think it's more of a thing where 40 layers of nested divs "feel" wrong, but they really have no real usability impacts on AT.

That being said you can [use SVG as a background image](https://codepen.io/netsi1964/pen/HGJms). And if the source to it is inlined as in the CodePen example I just linked users can still override the `background-image` and its embedded `fill` to change its appearance.

tvararu:
Ok let's dial it back a bit. 😄 

@edwardhorsford and I have had previous conversations about attempting a "constrained mode" design which does not allow free-text submission, like Chosen/Select2/other autocompletes. I feel this is out of the scope of this task, which is to provide a way to display results when no text has been entered.

The design change is small enough that I feel pretty confident about including this behaviour as a non-default option, an escape hatch to get this heavily requested feature, until we can scope in user researching "constrained mode" (which needs a better name).

With regards to forking, I think that's completely the wrong answer for this type of feature. I've seen it mentioned before (with regards to not offering a `placeholder` option because we disagree with it, and asking users to "fork if they want it") and I think this is wrong for a number of reasons. The two most important are:

- Forking divides the community
- Forking means duplication of work

Forking (when it entails creating a competing open source product) is a last resort to an unresponsive project, or a VERY significant change in direction.

Should our `options` start clashing between them and significant mutations of the UI occurring as a result of their application, I propose a different solution: exposing multiple components.

```js
accessibleAutocomplete({})

accessibleAutocomplete.constrainedMode({}) // Better name please

accessibleAutocomplete.experimentalMode({}) // Features we haven't user tested but are requested?

accessibleAutocomplete.whatever({})
```

This way they can all piggy back off of the same solid foundation.

---

TL;DR I think we should clean this up and ship it, and come back to the full "constrained mode" later.

src/autocomplete.css
edwardhorsford: Presumably both `.button--default` and `button--primary` could extend `.button` if we wanted.

tvararu: @edwardhorsford I think a modifier class isn't meant to live on its own like that, so I don't think it's necessary, but yes you can do it.

webpack.config.babel.js
nickcolley: Inline SVG is better for users as it does not require an additional request, you could also base64 inline into background: url() too

joelanman:
Updated to use SVG instead of PNG - so far I've tested in Chrome

Still to think about (that I can recall):

1. change css to not rely on specificity
2. use minLength instead of new parameter

I'm still not sure about 2 - it just doesn't seem obvious to me that changing minLength to 0 would:
- change appearance to a dropdown
- display all values on click as opposed to other events (pageLoad, focus) 

minLength is about how many characters you type before options appear. 0 just doesn't feel like a clear value considering that paradigm. But could just be me.

joelanman:
Now using the css approach suggested by @tvararu - can you check I've understood correctly? Not used this approach before

src/dropdown-arrow-down.js
```diff
@@ -0,0 +1,13 @@
+import { createElement, Component } from 'preact' /** @jsx createElement */
+
+export default class DropdownArrowDown extends Component {
+  render () {
+    return (
+      <svg version='1.1' xmlns='http://www.w3.org/2000/svg' class='dropdown-arrow-down'>
```
tvararu: You need `className=` here instead of `class=`.

src/dropdown-arrow-down.js
```diff
@@ -0,0 +1,13 @@
+import { createElement, Component } from 'preact' /** @jsx createElement */
+
+export default class DropdownArrowDown extends Component {
```
tvararu: Is `IconArrowDown` a better name as it isn't dropdown specific on its own?

src/dropdown-arrow-down.js
tvararu: Note that the `className` here includes positioning / layout styling:

```css
position: absolute;
right: 8px;
width: 24px;
height: 24px;
top: 10px;
```

This component would be more reusable if the layout styling was specified from its parent, so instead:

```js
<svg version='1.1' xmlns='http://www.w3.org/2000/svg' className={ this.props.className }>
```

And when you call it:

```js
<DropdownArrowDown className={ dropdownArrowClassName } />
```

tvararu: For the className itself, we should use a BEMmy one to keep with the other ones:

```js
const dropdownArrowClassName `${cssNamespace}__dropdown-arrow`
```

Which is `.autocomplete__dropdown-arrow` by default.

src/autocomplete.js
```diff
@@ -355,6 +361,7 @@ export default class Autocomplete extends Component {
     const inputClassName = `${cssNamespace}__input`
     const componentIsFocused = focused !== null
     const inputModFocused = componentIsFocused ? ` ${inputClassName}--focused` : ''
+    const inputModType = this.props.showAllValues ? ` ${inputClassName}--show-all-values` : ` ${inputClassName}--default`
```
tvararu: 👍

src/autocomplete.js
```diff
@@ -51,6 +52,7 @@ export default class Autocomplete extends Component {
     onConfirm: () => {},
     confirmOnBlur: true,
     showNoOptionsFound: true,
+    showAllValues: false,
```
tvararu: We need a little `README` blurb documenting this.

src/dropdown-arrow-down.js
joelanman: I think I'd rather keep dropdown as that's the specific context it's designed for - a more generic down icon (say for sorting) might be different

tvararu: Sound. 👍

tvararu:
Closing this in favour of https://github.com/alphagov/accessible-autocomplete/pull/176.

Issue #163, opened 2017-06-07
enhanceSelectElement always sets autoselect to true


joelanman:
`opts.autoselect = opts.autoselect || true`

should be 

`if (opts.autoselect === undefined) opts.autoselect = true`

I think

tvararu:
Yep, good catch!

joelanman:
just looking into this - why are we setting autoselect to true in `enhanceSelectElement`?

tvararu:
@joelanman it makes it more similar to a real `<select>` in that it doesn't allow you to get away as easily without picking an option.

joelanman:
not sure about that - most selects would have '-choose an option-' first value. Also this means you get different behaviour just because you're enhancing a select - thats a bit unexpected.

tvararu:
That's a fair point and I do agree with it, @edwardhorsford may remember the exact reasoning (we were also considering making `autocomplete: true` the real default at one point)

edwardhorsford:
Reasoning:

We're setting the default to the most sensible option in each case. For enhanced inputs we need users to be able to type anything, and not have to pick from the list. For selects, users *must* pick from the list. So the most sensible option is different for each.

The challenge here is to have consistent options, but also an easy way for us to provide 'recommended' sets of options. So far we've done this by having defaults that offer our recommended options.

My concern with changing this default is that we'll see most users of this component not using autoselect- a feature we've found to be very beneficial. I'm very open to alternate suggestions that achieve this though, and agree the inconsistency is awkward.

joelanman:
this was fixed by #167

PR #167, opened 2017-06-08
Fix autoselect in enhanceSelectElement

joelanman:
this fixes the default value check in `enhanceSelectElement`.

As an aside, I don't think the default should actually be set to true here, it's not set when using accessibleAutocomplete normally so I found it unexpected behaviour.


tvararu:
Relates to #163.

tvararu:
Thanks! 🌈

PR #173, opened 2017-06-26
Add explicit dependency on webpack-dev-middleware

### LINKED ISSUE #172: npm run dev is failing with npm install, working with yarn
Sorry this is a pretty vague issue.

To recreate:

1. Clone from scratch
2. npm install
3. npm run dev

Instead of running normally, it opens at `http://localhost:8080/undefined`, and none of the assets (css, js) load, instead they 404.

If I use yarn instead for step 2 above, everything works fine, so I imagine this is an issue with a package that has been updated.

My node version is 6.10.3, npm is 5.0.3

tvararu:
`webpack-dev-server` has a transitive dependency on `webpack-dev-middleware@^1.0.11`, but `1.11.0` breaks our development server for reasons unknown. (Might be this: https://github.com/webpack/webpack-dev-middleware/pull/187)

`1.10.2` is a version that works, and strictly specifying it in `package.json` will get both yarn and npm to respect it when resolving dependencies.

Fixes #172.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6898 @@
+{
+  "name": "accessible-autocomplete",
+  "version": "1.3.1",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.0.9",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+      "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
+      "dev": true
+    },
+    "accepts": {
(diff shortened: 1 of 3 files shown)


tvararu:
@joelanman can you please confirm if this fixes your issue? 🌈

joelanman:
I get no WARN any more from `npm install`, and `npm run dev` works - however it still opens `http://localhost:8080/undefined` instead of `http://localhost:8080` which seems odd

tvararu:
@joelanman cool! I’m happy to merge this for now and come back to investigate that later, as it seems harmless.

PR #174, opened 2017-06-28
Min length = 0 changes behaviour to dropdown

joelanman:
Based on #158 this uses minLength=0 instead of a new parameter


tvararu:
You're right, this is weird. I'm going to close this, let's instead go with `showAllValues`.

PR #176, opened 2017-06-28
Show all values 2 - show all values on click (like dropdown)

joelanman:
replaces #158 so we can get a PR app on heroku


joelanman:
bug: pressing space does activate the dropdown, but only by accident as ' ' matches some options. Space (and down arrow?) should open the dropdown the same as a click does.

joelanman:
I've added README update, and fixed className.

I've had a bash at handling space. Do you think it ought to be handled in `handleKeyDown` instead?

joelanman:
I added code to trigger the dropdown on space and arrow down. I think this is ready to go.

One thing - visually, you get an indication you can view all values because it looks like a dropdown. How can we communicate that to screen reader users?

joelanman:
@tvararu I've added the code you requested

joelanman:
@tvararu ready to go I think

carolinegreen:
@joelanman That's published now to version 1.4.0

joelanman:
@carolinegreen thanks!

PR #180, opened 2017-07-03
Redirect keypresses on an option to input

### LINKED ISSUE #179: If you type, then press down, then type, your first char doesn't appear
1. Start typing, eg 'ge'
2. Press down, this highlights an option in the list
3. Continue typing 'rm'

Expected: 'germ' is now in the input
Actual result: 'gem' is now in the input

The first character in step 3 seems to just return focus to the input, it doesn't enter the actual character.

tvararu:
Fixes #179.

This issue seems to have appeared when we updated to `preact@8.1.0` as part of [v1.0.6](https://github.com/alphagov/accessible-autocomplete/blob/master/CHANGELOG.md#106---2017-05-17).

I've added a test to hopefully prevent a regression in the future.

This is still broken in IE on the React version, but not on the Preact version.


carolinegreen:
👍

PR #183, opened 2017-07-06
v1.4.1

tvararu:
I've also updated the CONTRIBUTING as I realised the previously documented way of providing a message to `npm version` wasn't making them display as pretty as they can.


carolinegreen:
Looks good as far as I can tell.

PR #186, opened 2017-07-07
Disable hint

### LINKED ISSUE #120: Add option to disable `<Hint />`
We provide options for most other features, so maybe we should for this as well if people request it.

carolinegreen:
Allow users to hide the hint when using autoselect.  Closes #120.

With hint:

screen shot 2017-07-07 at 14 41 06 1

Without:

screen shot 2017-07-07 at 14 41 21


edwardhorsford:
Can you include before and after screenshots?

carolinegreen:
@edwardhorsford Yes good point.

tvararu:
Should update this with some tests.

tvararu:
The tests are failing because [IE9 doesn't have pointer events](http://caniuse.com/#search=pointer%20events); we need to wrap the tests to only run if `isIE9` (the variable is already declared but commented at the top of the tests).

tlwr:

PR #187, opened 2017-07-07
Add ability to prevent a flash of unstyled content

carolinegreen:
Add js-enabled CSS for the select which allows users to stop fouc if they add js-enabled to the html element.  Also update form example to reflect this.


edwardhorsford:
What happens if js is enabled but fails to load successfully?

tvararu:
@edwardhorsford With this, if JavaScript is enabled but it the autocomplete fails to render (connection problems, JS errors), the `<select>` will look like a prefilled input field:

screen shot 2017-07-07 at 14 58 04

Which when clicked, is actually a `<select>` element:

screen shot 2017-07-07 at 14 58 08

This isn't ideal, but not broken either.

Having the caret from `showAllValues` that Joe built could potentially help with this; is that something worth trying out?

nickcolley:
We'd have to create an error boundary around this component that can apply a class to override this.

e.g.

1. Add js-enabled in the <head>
2. Try to init component but wrap it in try {} catch {}
3. If it fails in the try {} catch {} add 'is-init-failed' class to element

Mind dump so might be a more sane way, on GOV.UK we just do a general check at the end of the page.

edwardhorsford:
An idea I had previously (which would help in most, but not all cases) is to hide the input on page-load, but display it using css transition after 2 seconds. If js fails to run, the input will become visible, but in most cases (assuming js loads in under 2 seconds), we'll just see the new autocomplete. The advantage of this method is it feels more robust - but at the downside that slower browsers or those not supporting css transitions will still get a FOUC.

tlwr:

Issue #96, opened 2017-05-03
Internationalisation


tvararu:
Things to do:

- [x] Allow customisation of the `aria-live` region
- [x] Allow customisation of the "No results found" template

Possibly others.

Possible API:

```js
AccessibleTypeahead.enhanceSelectElement({
  selectElement: document.querySelector('select'),
  lang: 'en' // Implicit default
})

AccessibleTypeahead.enhanceSelectElement({
  selectElement: document.querySelector('select'),
  lang: 'cy' // templates: AccessibleTypeahead.welshStrings
})

AccessibleTypeahead.enhanceSelectElement({
  selectElement: document.querySelector('select'),
  templates: {
    notFound: 'Nu s-a gasit'
  }
})
```

tvararu:
This has been partially done by https://github.com/alphagov/accessible-autocomplete/pull/190.

I'm keeping this open as I'd like to either document in the README or add a built-in option to quickly switch the language to Welsh, as this would be very useful for UK government departments.

tvararu:
This can be closed as there should be enough support for it.

PR #189, opened 2017-07-14
Introduce the ability to translate the texts and change the dropdown icon

sventschui:
Introduces the ability to translate texts and change the dropdown icon.

There are no breaking changes. The component behaves exactly like it did before when not using the newly introduces props.

Please see the changes in README for further information.


tvararu:
Excellent work, @sventschui! 👏 

This relates to https://github.com/alphagov/accessible-autocomplete/issues/96.

One issue I can see is the fact that this requires end users to know about React and provide functions that return React elements. We sell the plain JS API as the primary way of using the autocomplete as we expect this component to be used mainly in Government where (P)React isn't popular today.

To address this, you can change the various translation template functions to return strings:

```js
tQueryTooShort: (minQueryLength) => `Type in ${minQueryLength} or more characters for results.`,
tNoResults: () => 'No search results.',
tSelectedOption: (selectedOption, length) => `${selectedOption} (1 of ${length}) is selected.`,
```

And update their call sites to wrap them in the respective `<span>`s and what not.

This doesn't work as nicely for the `DropdownArrow`, so I propose we move that to a separate pull request to discuss the best way to address that.

sventschui:
I'll open another PR for the separate parts

PR #194, opened 2017-07-17
Support using down arrows in IE9-11

### LINKED ISSUE #193: IE10 - using arrows to navigate list selects item by mistake
Using IE10 (checked on Win 7 and Win 8) when typing in a country (eg Fr) and then using the down arrow key to navigate the list. 
When the key is pressed after the highlight is on the first item in the list, the second item in the list is automatically selected.

arrow key ie10

carolinegreen:
Change onBlur to onFocusOut to support IE9-11.  onFocusOut isn't ideal as React onFocusOut doesn't bubble currently, however onBlur is not supported at all on IE9, 10 and 11.

Fixes #193


tvararu:
The downside to doing this is that now the React version fires lots of warnings. There doesn't seem to be a way to fix this currently; this issue has more relevant discussion: https://github.com/facebook/react/issues/3751

tvararu:
Published in v1.4.2!

PR #190, opened 2017-07-17
Introduces the ability to translate texts

sventschui:
Introduces the ability to translate texts

Follow up to #189, closes #96 

There are no breaking changes. The component behaves exactly like it did before when not using the newly introduces props.

Please see the changes in README for further information.

PR #196, opened 2017-07-18
Introduce the ability to translate texts

tvararu:
Rebase of #190. Also added a CHANGELOG entry with proper attribution.

Reopening this also means it can run on the full suite of Sauce Labs browser tests.


tvararu:
Published in v1.5.0!

PR #209, opened 2017-07-25
Fixed form example

svinkle:
#208 broke the example; should have tested the change locally before submitting… apologies.


tvararu:
Don't sweat it, this highlights a gap in our testing, as I didn't try it locally either!

PR #214, opened 2017-08-10
Revert fix ios custom click

### LINKED ISSUE #177: Scrolling list on touch device selects first item touched
Look at https://accessible-autocomplete-pr-176.herokuapp.com/examples/ **on a mobile or touch device**

**Steps to recreate bug**
1. Pick any of the autocomplete examples and type "a"
2. Scroll this list, noting the country that you touched to start the scroll
3. As you reach the end of the scrollable area, let go

**Actual Result:**
The country that you touched to begin the scroll will be selected

**Expected result:**
 I would be able to carry on scrolling until the end of the list

**Impact**
- Users will be unable to select the correct item from the list
- Once an item is selected it's hard for a user to understand how to see the list again (they need to delete the current selection using backspace) 
- The `{ showAllValues: true }` option will be unusable on a touch device as scrolling is the only way to interact with it

Only tested on iOS Safari but I would imagine Android Chrome would have the same problem

danmitchell-:
Reverting [this](https://github.com/alphagov/accessible-autocomplete/pull/85) which has minimal usage right now but fixes this larger/more important issue #177.

A straight Git revert isn't possible given its place in time and a lot of code has changed since then, even the file name.


edwardhorsford:
@danmitchell - can you give some more context on this? What is the impact of removing it? what does it do?

danmitchell-:
@edwardhorsford the change I am reverting is https://github.com/alphagov/accessible-autocomplete/pull/85

This feature is rarely used at present and being able to scroll on mobile is much more important right now as it's hindering adoption. I have created an issue for custom select on mobile but i'm really struggling to replicate it anyway.

edwardhorsford:
Thanks @danmitchell-. I don't understand what the feature does though - can you explain it?

danmitchell-:
I'm not 100% sure @edwardhorsford, hence the reason i'm struggling to replicate. I think it's something todo with https://github.com/alphagov/openregister-picker-engine#optionsadditionalentries

edwardhorsford:
Thanks @danmitchell- it sounds like a good thing to revert, but it would be good to know *what* we're losing, and open an issue to fix it. How do we know that has very low usage? Is it because it's an iOS only thing?

tvararu:
@edwardhorsford this was a fix for custom results (see end of [examples page](https://alphagov.github.io/accessible-autocomplete/examples/)) not being tappable on iOS devices when VoiceOver is turned on. Unfortunately, the fix breaks scrolling on iOS when VoiceOver is turned off…

I agree that this should be reverted until a better solution is found for the tapping bug.

src/autocomplete.js
```diff
@@ -21,7 +21,7 @@ const hasPointerEvents = (() => {
 })()
 
 function isIosDevice () {
-  return !!(navigator.userAgent.match(/(iPod|iPhone|iPad)/g) && navigator.userAgent.match(/AppleWebKit/g))
```
tvararu: The `!!` here coerces the value to a boolean, if you don't do it then you'll return the value of the `.match` function which is a string, and could be confusing.

```js
> isIosDevice()
'AppleWebKit'
```

vs

```js
> isIosDevice()
true
```

danmitchell-: Thanks @tvararu ! Will remove this change :)

PR #217, opened 2017-08-18
fix(a11y): update element roles and aria expanded to be set if false

tobias-g:
Hi

We noticed some aria issues and role issues our [aXe](https://github.com/dequelabs/axe-core) tests came up with.

`aria-expanded={menuOpen}` that was on the input when it was false because of the binding setup the attribute was not added at all when expected behaviour is when false the attribute is present and displayed as `aria-expanded="false"`.

Also moved the aria-expanded attribute and combobox role to the wrapper elements, then added role textbox to the input as aXe was not happy with this either.

Hope that all makes sense cheers 🍺


test/functional/index.js
```diff
@@ -68,6 +68,26 @@ describe('Autocomplete', () => {
         expect(scratch.innerHTML).to.contain('class="bob__input')
         expect(scratch.innerHTML).to.contain('class="bob__menu')
       })
+
+      it('renders with the correct aria attributes', () => {
+        render(<Autocomplete required />, scratch)
+
+        console.log(scratch)
```
tvararu: You forgot a `console.log` here. 😄

tobias-g: Ha indeed 😱 my bad, fixed that now 👍

tobias-g:
@tvararu I pushed updates to the branch to remove my console.log. Anything else to update drop me another message ta :)

danmitchell-:
@tobias-g if you can rebase with master then I can merge this for you 👍

tobias-g:
Thanks @danmitchell- rebase done 😁

PR #226, opened 2017-08-31
fix: selection and timeout race condition

tobias-g:
We noticed an intermittent issue where a user would select an item from the typeahead and instead of filling in the input with the value and closing the menu below the input value would not change and the menu remains open requiring the user to click again to make the selection.

This occurred more frequently on IE than other browsers though it did seem affect all browsers.

Some digging lead me to believe it was related to `pollInputElement`. Here's the steps that look like they are the cause of this race condition:

- User selects an item and this ends up calling `handleOptionClick`
- `handleOptionClick` updates the state to set the selected value and close the menu
- Here render should be called and the input value updated in the DOM and menu hidden however in some occurances... `getDirectInputChanges` is called via `pollInputElement` before render is called
- `getDirectInputChanges` grabs the value from the input (because render has yet to be called its not the selected value in the state)
- `handleInputChange` is called because state value and input value does not match
- Now the menu state is updated to be open and selected value in state is updated to input value
- Render gets called and user ends up seeing open menu and their selection not selected

My React know how is pretty green so input on whether this makes sense or the fix I suggest is not optimal let me know. Also really sorry but not sure on best way to unit test this fix. It may make the webdriver tests more stable though?

Fix I added was to forceUpdate to get render to call whenever `handleOptionClick` is called.

Cheers 🍺


samtsai:
@tobias-g having similar issues but what's worse is it seems to happen all the time in certain parts of our app. Tested the areas with this PR and it's working great!

danmitchell-:
Thanks for this @tobias-g, lets get #217 merged first and then we can look at this one and #226 👍

tobias-g:
@danmitchell- rebased this one so good for you to take a look at now, thanks.

samtsai:
Hi @tobias-g @danmitchell- can this one be moved forward again?

danmitchell-:
@tobias-g @samtsai sorry guys, back on this now. Would you mind rebasing again @tobias-g and then i'll merge it?

samtsai:
@tobias-g @danmitchell- I created a fork and PR of a rebased version with an additional "fix" see #238

samtsai:
This can be closed now that #238 is merged, thanks @tobias-g for the actual work!

danmitchell-:
Thanks for this guys!

tobias-g:
Struggled to get time to rebase this so thanks @samtsai for getting it sorted :)

PR #228, opened 2017-09-08
fix(ie): tab focus fix for IE on SVG element

tobias-g:
When `showAllValues` is true and the SVG for the dropdown renders IE will see this element as focusable (other browsers we tested ignore the SVG by default). PR fixes this by adding focusable=false so a user does not have to tab twice to move out of the typeahead.

Had to use focusable=false instead of tabindex see: https://stackoverflow.com/questions/18646111/disable-onfocus-event-for-svg-element

Added couple of basic tests to go with it.

Cheers 🍺


tobias-g:
@danmitchell- Cheers for the approve. I've rebased so its good to merge though I see theres a few other PRs in the works so will happily rebase again if those take priority 👍

Issue #230, opened 2017-09-12
The example page is broken


aduggin:
I'm only seeing drop downs on this page - the autocomplete isn't being rendered:
https://alphagov.github.io/accessible-autocomplete/examples/form.html

joelanman:
this is broken again

danmitchell-:
Looks okay now. Was caused by the v8-lazy-parse-webpack-plugin package which was redundant anyway.

Issue #213, opened 2017-08-08
Provide a way to progressively enhance an input


edwardhorsford:
The autocomplete needs to be able to progressively enhance from an input as well as a select.

Where the options are known and reasonably few in number (countries), select is fine. For a large number of options or unknown options, we need an input. Non js would submit what the user enters - the server can then do their own checks, or search or whatever.

Related to #95. Also the opposite of #198

trevorkapswarah:
Hi,

This is an issue we have faced too as we need an existing input element to be bound to the autocomplete functionality as the input element can serve more than one purpose.

matthewford:
We're using it to enhance an input on the trade tariff's search: https://www.trade-tariff.service.gov.uk/sections?currency=EUR

PR #232, opened 2017-09-12
Fix intermittent JS error on example pages

danmitchell-:
We get an intermittent JS error `__webpack_exports__ is not defined` on all example pages which caused the `accessibleAutocomplete` not to load.

Have a look around the issue is coming from `v8-lazy-parse-webpack-plugin` and we don't need it anymore - https://github.com/webpack/webpack/issues/3980 && https://github.com/zeit/hyper/issues/1418

Using `yarn dev` works for some reason but `npm run dev` doesnt. Very odd.

This will fix https://github.com/alphagov/accessible-autocomplete/issues/230


edwardhorsford:
Probably good evidence for us to try to do #213 at some point too - right now most examples aren't there as they're not enhancing anything. We should aim that we're always enhancing *something*
screen shot 2017-09-12 at 13 06 01

PR #233, opened 2017-09-12
Update author

danmitchell-:


package.json
```diff
@@ -5,7 +5,7 @@
   "style": "dist/accessible-autocomplete.min.css",
   "description": "An autocomplete component, built to be accessible.",
   "repository": "alphagov/accessible-autocomplete",
-  "author": "Theodor Vararu <theodor.vararu@digital.cabinet-office.gov.uk>",
+  "author": "Government Digital Service developers (https://gds.blog.gov.uk/)",
```
edwardhorsford: Could I suggest "Government Digital Service" and link [here](https://www.gov.uk/government/organisations/government-digital-service) as that matches what we put in the footer of GOV.UK?

edwardhorsford:
Thanks Dan, looks good 👍

danmitchell-:
🤔 I wonder if this failure is something todo with https://github.com/alphagov/accessible-autocomplete/pull/232

PR #243, opened 2017-09-28
dispatch 'confirm' event

revilossor:
just a little change so the autocomplete__input element will dispatch a change event when an option is chosen from the list.

this event will also be dispatched from enhanced select elements, so you dont have to wait for enhancement to finish before setting up listeners on new elements


revilossor:
Bump?

36degrees:
Hi @revilossor,

Thanks for raising this, and sorry it's taken so long to get a response.

I just wanted to check if you were aware of the [`onConfirm`](https://github.com/alphagov/accessible-autocomplete/#onconfirm-default---) option – is this something that would meet your needs?

The autocomplete is quite complex already, so if we can avoid having two ways of handling events I'd be keen to do so.

If this is something that you still think we should add, to get this merged I think we would need:

- this branch rebased against master and any conflicts resolved
- tests added that cover the new functionality
- documentation that covers the new functionality
- the commit history tidied up a little

Unfortunately our team don’t have the time at the minute to be able to help with this, so we'd be looking for you or the community to pick that up.

I’m conscious this PR is quite old and hasn’t had any activity for a while – we’ll keep it open for another couple of weeks, and if there’s been no activity we’ll close it and turn it into an issue.

Thanks.

PR #254, opened 2018-02-15
Redirect /dist/ assets to route

nickcolley:
This is so that we can have github pages and the dev server work alongside.
Since github pages relies on the structure of the repo, whereas webpack-dev-server does not.

Fixes #253

Issue #253, opened 2018-02-15
`npm run dev` returns 404s for accessible-autocomplete


codekiln:
```
$ node -v
v9.5.0

$ npm -v
5.6.0
```

OS X 10.12.5

When I clone the repo, check out the tip of the master branch (64b07d76b), run `npm install` and `npm run dev`, I get 404s for accessible-autocomplete, and the examples are currently not working, with "Uncaught ReferenceError: accessibleAutocomplete is not defined" appearing in the console.

![dev tools network tab](https://user-images.githubusercontent.com/140930/36252748-9f06d3f4-1213-11e8-9396-1648116a331b.png)

![dev tools console](https://user-images.githubusercontent.com/140930/36252945-32668798-1214-11e8-8a33-40984b56f499.png)

nickcolley:
Hey @codekiln :)

I think I know why this is broken, PR incoming.

Thanks for raising!

PR #257, opened 2018-02-20
Pin Firefox version for SauceLabs tests

hannalaakso:
Part fix for failing Travis build reported in
https://github.com/alphagov/accessible-autocomplete/issues/253

We hadn't tied SauceLabs to a version of Firefox. We haven't done
active development on the repo for a bit so it looks like the latest FF is not
compatible.

Tested as part of this PR: https://github.com/alphagov/accessible-autocomplete/pull/254

I wonder if we should specify Chrome version as well?


36degrees:
👏

PR #272, opened 2018-05-26
Make sure defaultValue is rendered and apply template to it

iensu:
This PR does two things:
1. It enables using the `defaultValue` prop when working with a list of objects. This is done by applying the `inputValueTemplate` to the `defaultValue` when setting the initial query.
2. It updates the `query` and `options` if the `defaultValue` goes from `undefined` to a defined state. This enables you to pass a default value to the Autocomplete component when the data is loaded asynchronously, without having to trigger a surrounding re-render.


iensu:
Your CI seems to be a bit broken judging from the logs. :(

tlwr:

PR #278, opened 2018-07-04
fix fill-rule attribute, JSX required camel case fillRule

adamsilverstein:
### Changes

This should fix a console warning I see when using the `Autocomplete` component in our React (vs 16) project. The attribute appears misnamed, JSX requires a camel case equivalent, see https://reactjs.org/docs/dom-elements.html#all-supported-svg-attributes

### Screenshot
Before this change I get a console warning:

### DIFF
--- a/src/dropdown-arrow-down.js
+++ b/src/dropdown-arrow-down.js
@@ -2,7 +2,7 @@ import { createElement } from 'preact' /** @jsx createElement */
 
 const DropdownArrowDown = ({ className }) => (
   <svg version='1.1' xmlns='http://www.w3.org/2000/svg' className={className} focusable='false'>
-    <g stroke='none' fill='none' fill-rule='evenodd'>
+    <g stroke='none' fill='none' fillRule='evenodd'>
       <polygon fill='#000000' points='0 0 22 0 11 17' />
     </g>
   </svg>


36degrees:
Hi Adam,

Thanks for raising this – sorry for the delay in replying. This doesn't seem to be working for the Preact implementation – when running the development app using `npm run dev` I'm seeing `fillRule` on the element in the inspector, so it's not being templated back to kebab-case. I think this might be related to https://github.com/developit/preact/issues/813. 

Is that something you'd be able to look at fixing? If not I can get this added to our backlog and we can try and find an approach that works in both cases.

adamsilverstein:
@36degrees Interesting! I will take a look when I get a chance (which may be a bit), I'd love to help make sure it works in Preact as well!

adamsilverstein:
@36degrees How can I reproduce the errors you were seeing? I ran `npm run dev` and don't see any errors or console warnings - do I need to enable Preact explicitly?

36degrees:
I'm not seeing errors, per se – but the attributes on the SVG aren't quite right.

Running the dev server (`npm run dev`), when I inspect the dropdown arrow in Chrome's developer tools, I can see the `<g>` element has the `fillRule` attribute, but it should be [`fill-rule`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule):

```
<g stroke="none" fill="none" fillRule="evenodd"><polygon fill="#000000" points="0 0 22 0 11 17"></polygon></g>
```

screen shot 2018-09-03 at 16 57 08

If I edit the attribute to `fill-rule` in Chrome's dev tools, I can see it then appears in the element's styles (right-hand pane):

screen shot 2018-09-03 at 16 57 12

adamsilverstein:
Thanks for the details, I'll take another look.

36degrees:
Hi @adamsilverstein,

Did you manage to get anywhere digging into this?

I’m conscious this PR is quite old and hasn’t had any activity for a while – we’ll keep it open for another couple of weeks, and if there’s been no activity we’ll close it and turn it into an issue, as I don't think this is the right fix currently.

Thanks again for your help.

adamsilverstein:
Hey @36degrees unfortunately i've been heads down in project work. I think opening a matching issue to track this is fine, thanks for doing that. Can you please cc: me on the issue so I can follow along with any updates?

I **am** still trying to merge `accessible-autocomplete` for use in the new editor for WordPress, see https://github.com/WordPress/gutenberg/pull/7385 - this is where I came across the issue originally.

adamsilverstein:
@colinrotherham was this fixed in https://github.com/alphagov/accessible-autocomplete/pull/315? If not, can you give me any tips on proceeding here, I'm a bit nervous about jumping in without muich experience in the project.

Thanks!

nickcolley:
Since there's not been activity for a while I'm going to close this out but if you do any more investigation feel free to raise a new issue for us to investigate, thanks a lot for helping so far.

PR #287, opened 2018-07-27
A Fix for Issue #286 — Spacebar input not registered when 'No results found'

### LINKED ISSUE #286: Spacebar input not registered when 'No results found'
When the autocomplete is configured with ``` { showAllValues: true }``` and a string is entered which does not create a match, but includes a space, the first spacebar key-press is not registered.

For example, take the use-case of an autocomplete/typeahead which allows users to also provide inputs which do not match the existing list of options — perhaps to store the users option against the list at a later point. 

When a user attempts to enter 'Petrograd and Leningrad' into the ``` { showAllValues: true }``` example:
image

By the time they have entered `Pet` there are no results left:
image

The user continues to type `rograd` without any problems: 
image

But should they continue to type `rograd and Leningrad` the following is registered:
image

In order to overcome this issue, the user is forced to double hit space for each attempt to add a single space. The first space not being registered as it is used to open the option to show all values.

AdenFraser:
This is a proposed fix for Issue #286 

The handleSpace method is used to either select a focused option when one exists, or open the full list of options when it is not open. However, there is an edge-case where the user is attempting to enter a string which includes a space, which is not in the list. The space-key entry is omitted on first try, resulting in a malformed string.

This fix simple checks whether there are available results/'No results found' message (or equivalent) is showing — and if that is the case, there is no need to re-attempt to open the menu, thus allowing the user to input their string as expected.

Fixes #286


src/autocomplete.js
```diff
@@ -318,7 +318,7 @@ export default class Autocomplete extends Component {
 
   handleSpace (event) {
     // if not open, open
-    if (this.props.showAllValues && this.state.menuOpen === false) {
+    if (this.props.showAllValues && this.state.menuOpen === false && this.state.options.length) {
```
36degrees: I think we can use [`menuIsVisible`](https://github.com/alphagov/accessible-autocomplete/blob/master/src/autocomplete.js#L423) instead of `menuOpen` here, which I think better captures the intent and avoids adding another expression?

36degrees:
Thanks for this Aden! I've had a little play locally and your fix does solve the problem, but I've got a suggestion I think makes it slightly clearer.

src/autocomplete.js
AdenFraser: That makes a lot more sense. Sorry am new to this package and codebase, didn't see the menuIsVisible method!

I'll get that amended and pushed.

36degrees: No worries – I'm still getting to grips with it myself!

AdenFraser: Actually I'm not too sure how best to refactor this, as the menuIsVisible const is insider the render()?

Making menuIsVisible available from handleSpace will also require ```const showNoOptionsFound = this.props.showNoOptionsFound &&
      inputFocused && noOptionsAvailable && queryNotEmpty && queryLongEnough``` from the render() too? Which then relies on the latter four also being available.

Unless I'm overlooking something — it's mighty hot in our office! 😅💦☀️

36degrees: Sorry Aden, you're quite right (I told you I was still getting to grips with it myself!)

From talking to @joelanman who originally added this feature, the intent is that when you focus the element using the keyboard (i.e. by tabbing to it) and press space it should open the menu and display all results (mimicking behaviour of native selects).

I've just tested your fix and whilst it corrects the behaviour that you're seeing (adding a space whilst seeing no results triggers the menu), it _doesn't_ quite work in that scenario –  `this.state.options.length` is 0 if there is no query, so this handler doesn't kick in and rather than opening the menu it just searches for all items with a space in the name.

I _think_ it makes sense to open the menu _if_ it is not already open and there is not already a query – which would work in both cases I think? In which case changing the line to…

```
if (this.props.showAllValues && this.state.menuOpen === false && this.state.query != '') {
```

might make sense? I've done some initial testing and it seems to work well, but it'd be good if you could double-check that solves your problem.

AdenFraser: Hi, gave that a go but that didn't work — although I think we actually just need ```this.state.query === ''```, since we only wish the the options list to appear when handling space input initially, not with latter space input.

I've tested it with both the intended original scenario and against the issue I was facing and both now seem to be resolved.

Have rebuilt ```dist```, rebased and squashed and all pushed up.

Aden

Issue #271, opened 2018-05-18
Result count incorrectly read with JAWS and NVDA on Chrome.


edwardhorsford:
Bug report via twitter:
["Hi, I tested the autocomplete withJAWS & NVDA on Chrome. For the results list it keep saying 1of58 even when I am on 5th result. It should say 5 of 58. It's a minor bug."](https://twitter.com/DigitalA11y/status/996981916598353920)

Note this isn't a browser / AT combo we recommend testing in. Creating the issue though so it's recorded.

PRGfx:
This is because the option `tStatusSelectedOption` (or the `tSelectedOption` prop of Status) does not receive the currently selected index and returns `1 of __`. I think the autocomplete's state value `selected` should be available in the `Status` component as e.g. `selectedOptionIndex`.

PR #291, opened 2018-08-06
Pass selected option's index to Status; use in tStatusSelectedOption

PRGfx:
This pull-request exposes the currently selected option's index to the Status component to be used in the `tStatusSelectedOption` label and addresses #271.
Implementations currently setting the `tStatusSelectedOption` to a function without a third argument should not be concerned.


PRGfx:
I edited in my comment on the corresponding issue #271 that this behaviour I tried to fix, might in fact be intended as it is. Maybe someone could clarify on that.
I don't have a comparison right now, but at least chromevox already announced the item position on it's own before appending the "1 of n" status, which was slightly confusing and a reason for this PR.

PRGfx:
It could if it was possible to select multiple elements and the status would relate to the amount of items selected, but I think without the annotation of supporting multiselect, a status on how many items are selected would not be expected...

peterjaric:
Sorry for poking at this closed issue, but [examples/index.htm](https://github.com/alphagov/accessible-autocomplete/blob/master/examples/index.html) still uses the old "1 of ..." style:

        tStatusSelectedOption: function (selectedOption, length) {
          return selectedOption + ' (1 von ' + length + ') ist ausgewählt.'
        },

If you follow this example, you get different behavior between English and your localized version. Should this be fixed? I can make a PR for this little change, if needed.

nickcolley:
@peterjaric if I'm following correctly are you saying that the "1 of ..." format does not make sense when translated? We'd welcome a small pull request to make any translations more accurate in the examples. :)

peterjaric:
No, sorry, I mean that the new default English version is using the new `index` parameter (and adding one to it):

```
    tSelectedOption: (selectedOption, length, index) => `${selectedOption} ${index + 1} of ${length} is highlighted`,
```

But the example doesn't.

So if you go by the example and use the hardcoded string "1 av" (in Swedish), then you get one behavior in Swedish, it will say "1 av 8" if you're on the third option. If you switch to English (and use the defaults): it will say "3 of 8" instead.

I think the example should use the `index` parameter too.

nickcolley:
Ah I see what you mean, yes it looks out of date then.

If you wanted to contribute an update that'd be great, otherwise if you are busy could you please open a new issue so we can look to fix this in the future?

Issue #307, opened 2018-10-17
Bold text in results can't be clicked in IE11


colinrotherham:
This is an issue seen when using custom templates.

For an example see the Country and Territory autocomplete:
https://github.com/alphagov/govuk-country-and-territory-autocomplete

Each autocomplete result is shown in a list item:

```html
<li class="autocomplete__option" role="option" tabindex="-1">
  <strong>Guinea-Bissau</strong> (The Republic of Guinea-Bissau)
</li>
```

The `opts.templates.suggestion` function adds **\<strong\>** (in pink below) around each result:

![unclickable](https://user-images.githubusercontent.com/415517/47084392-42ae6a00-d20b-11e8-816b-796d0ea4ab0f.png)

**What's the bug?**
I've found two bugs so far in Internet Explorer 11:

1. The `<strong>` areas marked in pink are unclickable
2. When mousing over each `<strong>` the first item flickers "active"

Perhaps the event isn't bubbling up to the `<li>` item correctly?

**Video demonstration**
https://youtu.be/AEiblEaCY-A

Looking at the tests this is a known but unlogged issue:
https://github.com/alphagov/accessible-autocomplete/commit/9ad4eeccb7bf96b82a711d9405461fd81896bf3a#diff-cfce3706f03f2afd25608bfdae2bee4eR120

```js
if (isIE) {
  // FIXME: This feature works correctly on IE but testing it doesn't seem to work.
}
```

Thought I'd best get it logged if it helps with any design system considerations @nickcolley.

colinrotherham:
I've narrowed this down further and found two issues.

1. **IE11 gets stuck in a `mouseout` loop**, as seen below
Workaround: Only fire `mouseout` on the outer list `<ul>` not on every `<li>`.

![event loop](https://user-images.githubusercontent.com/415517/47713562-7f7a5800-dc32-11e8-8ab6-2e6dcdcbd28d.gif)

2. **By design, the result list closes when blurring away from the `<input>`**
There's a `mousedown` event preventer to fix this but it doesn't work in IE11.

Workaround: Remove “mousedown” blocking. Replace with short delay on results list close state to allow click to reach each result first.

[...]

PR #310, opened 2018-10-30
Fix mouse event issues in IE11

colinrotherham:
As discussed in https://github.com/alphagov/accessible-autocomplete/issues/307

Three main changes:

1. **Delay closing of results list to allow `click` to reach result item**
This is a better solution than blocking `mousedown` that works in IE11

2. **Run click method after `mousedown` event on iOS**
Workaround for [Safari mouse emulation behaviour](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html#//apple_ref/doc/uid/TP40006511-SW4)

3. **Fix looping `mouseout` event in IE11**
This was causing performance/flicker issues for our service

event loop


hannalaakso:
Hi @colinrotherham 

Thanks very much for this contribution 🙌 We'll review it at the beginning of next week and come back to you then.

I noticed that some of the tests are now failing. Would you know why that is? We can also take a look at fixing the tests if you prefer.

colinrotherham:
Hi @hannalaakso,

First failure was caused by `node-gyp` and Node.js 11 incompatibilities.
I've rolled Travis back down to LTS (Node.js 10), that fixed that part.

Next it looks like I've regressed re-focusing the input after an option is clicked:

```js
expect(browser.hasFocus(input)).to.equal(true)
```

There are so many events being fired it's a possible race between input blur (removes focus) and clicking an option (retains focus) but it's very hard to diagnose 😔

Take another look if you can? Wonder if @tvararu is still around?

tvararu:
@colinrotherham I can take a look maybe by weekend if I don't find time until then, but yeah! The eventing is very tricky.

colinrotherham:
Hi @hannalaakso @tvararu I've had to add a `{ clicked: index }` to the state.

If a "click" has taken place, we can then use this to move focus to the input without preventing events.

If you're happy with this new approach I'll start checking SauceLabs and some device testing.

examples/index.html
```diff
@@ -575,18 +575,26 @@ <h3>Translating texts</h3>
...
             is: (length === 1) ? 'ist' : 'sind'
           }
 
-          return `${length} ${words.result} ${words.is} verfügbar. ${contentSelectedOption}`
+          return length + ' ' + words.result + ' ' + words.is + ' verfügbar. ' + contentSelectedOption
         },
-        tNoResults: () => 'Keine Resultate'
+        tNoResults: function () {
+          return 'Keine Resultate'
+        }
```
hannalaakso: 👍 

Non-blocking: It would be good to note somewhere that script in `/examples` files is not transpiled and should therefore use pre-ES6 syntax. Maybe a new entry "Examples" for  [CONTRIBUTING.md](https://github.com/alphagov/accessible-autocomplete/blob/master/CONTRIBUTING.md)?

Issue #200, opened 2017-07-20
iOS Voiceover - empty input says '1 result available'


adamliptrot-oc:
When arriving at the autocomplete for the first time and moving from the label to the input, Voiceover announces '1 result available'.
The next navigable item is 'comma' which looks to be a comma inside a span tag.

![image](https://user-images.githubusercontent.com/14074579/28427120-146548e8-6d6d-11e7-9dd5-8b9784fbc94e.png)
(screenshots from Chrome to show code issue)

Not sure what the purpose of the comma is? Also looks that as you interact with the autocomplete, the comma is doubled.

![image](https://user-images.githubusercontent.com/14074579/28427251-7a0cb172-6d6d-11e7-80bb-3ac60b81c416.png)

The erroneous result count and presence of the comma is confusing for the user.

tvararu:
The 1 result is available sounds like a bug. It should say "No search results."

---

The comma is a hack. 😄 

Let's say you've typed in `unit` and have the following results:

- United States
- United Arab Emirates
- United Kingdom

The first time you finish typing `unit` you'll hear "3 results are available."

When you key in `e` and you change to `unite`, the number of results doesn't change. The `aria-status` region remains unchanged, users don't hear any feedback, so users think the autocomplete isn't searching anymore.

The way this was [previously solved](https://github.com/alphagov/accessible-autocomplete/pull/13) is by using a debounced `setTimeout` to clear the content of the aria status region. This is an approach used by other autocompletes, but I've found it to be buggy.

The comma hack by contrast (combined with `aria-atomic`) works, but may confuse the user. For what it's worth, only VoiceOver consistently pronounces it! NVDA or JAWS I found would sometimes simply ignore it.

While this particular quirk has not been user tested with screen reader users, from speaking with other accessibility developers the consensus was that it shouldn't be a barrier. SR users are used to hearing all kinds of weird stuff from their screen readers, whether it's mispronounced dates, "Governor UK," or punctuation in odd places. It's far more important to consistently report the count of results.

That being said, if you can think of a better way to fix the underlying issue, by all means we should use that instead. 👍

adamliptrot-oc:
The second comma seemed to appear when you interact with the autocomplete, but when trying to replicate it, it looks like it's always there.

tvararu:
@adamliptrot-oc cool, then yes that's a legitimate bug and we should look at it.

tvararu:
[...]

Issue #277, opened 2018-07-03
Autocomplete: consider to use `aria-posinset` and `aria-setsize`


afercia:
Testing a bit the autocomplete with Firefox and NVDA, it works pretty well (maybe it's the browser / screen reader combo where it works best) and it's great to see all the work that has been done here.

I've noticed something that could be improved when the suggestions list gets filtered. Seems Firefox and NVDA are not able to calculate and announce the new length of the list. In this case, I _think_ `aria-posinset` and `aria-setsize` could help.

In the screenshot below, when the list shows just 2 results, Firefox and NVDA still announce the initial length of the list (which was 165), thus the two options are announced as "Italy 1 of 165"  and "Mauritania 2 of 165", while they should be announced as "Italy 1 of 2"  and "Mauritania 2 of 2":

![screenshot 27](https://user-images.githubusercontent.com/1682452/42232772-c3c17e26-7eef-11e8-9920-ca4e1ce4bc27.png)

I haven't looked at the code, but if this was an array, I guess there's no interest in communicating the original position of the option within the initial 165 options so I guess [aria-posinset](https://www.w3.org/TR/wai-aria-1.1/#aria-posinset) should be the array item index and [aria-setisize](https://www.w3.org/TR/wai-aria-1.1/#aria-setsize) the array length.

Worth noting both properties are mentioned for the Listbox widget in the ARIA Authoring Practices: https://www.w3.org/TR/wai-aria-practices/#Listbox and allowed for the `option` role: https://www.w3.org/TR/wai-aria-1.1/#option

nickcolley:
Will be in the 2.0.0 release :)

PR #348, opened 2019-07-10
Refinements to address accessibility issues

markhunter27:
**_NOTE: This PR closed without merge - replaced by [this subsequent PR](https://github.com/alphagov/accessible-autocomplete/pull/355), which contains the same changes applied in a more structured manner. 
Comments and discussion here remain valid for traceability._**

The changes are focused on improvements in the following browser / screen reader combinations:

- Internet Explorer and JAWS
- Firefox and NVDA
- Safari and VoiceOver
- iOS Safari and VoiceOver 

..although we've also considered the NVDA and JAWS readers with each combination of Firefox, IE, and Chrome.

The changes are in part refinements to the markup of the component (type and placement of ARIA attributes), and in part changes to the detail and timing of ARIA live region updates ('status' updates in the terminology of the component).  

**Summary of refinements:** 

- Ensure immediate announcement of hint/status text when field receives focus.

      - Previously 'Select your country edit blank"
      - Now 'Select your country combobox collapsed editable blank. Type in 2 or more characters for results."

- On typing in a field with a 2 character minimum configuration (optional configuration)

      - Previously typing  'a' .. the typing echo reads 'a', and nothing else
      - Now reads 'a' .. then reads 'type in 2 or more characters for results'

- On finding no search results..

      - Previously, reads 'No search results dot' .. multiple times (up to 7 in some cases)
      - Now, reads 'No search results' a single time, and nothing else.

- On finding search results

       - Previously typing a few characters e.g. 'as' .. reads '8 results are available' multiple times.
          Selecting one of the search results then reads "8 results are available, UK, 1 of 8 is selected dot" - multiple times                 
       - Now, reads '8 results are available' a single time
          Selecting one of the search results reads '8 results are available, UK, 1 of 8 is highlighted' once. (period removed)

- Generally, no longer reads 'comma' or 'dot' on aria announce. Have refined the previous 'comma' approach (see [here](https://github.com/alphagov/accessible-autocomplete/issues/200#issuecomment-316755847) for original discussion). Now using a pair of aria live regions and flip-flopping status updates between. This change addresses the problem of announcing duplicate status messages (e.g. where number of results has not changed with new user input), and also the issue of repeated duplicate announcements that are caused by React virtual DOM behaviour.  See [here](https://github.com/nvaccess/nvda/issues/7996) for background inspiration.  
  
- Form fields are now visible in form overview modes (e.g. NVDA Insert + F7 , rotor in VoiceOver)

- (Related to the above) Form fields can now be selected directly using relevant form navigation key (e.g. key 'F' in NVDA)

- Reduced overly verbose selection announce in JAWS on IE, by removing role=alert

- Simplified aria live region 'div' elements (removed inner span element). Improves situation in JAWS on IE where aria live content would not otherwise be announced without first disabling 'virtual PC cursor' mode (which is on by default).

- Introduced a general event debounce on status update (aria live region update). This delays and limits frequency of aria live updates, and has the following benefits:

       - Improves situation on Mac Safari / VoiceOver where typing echo can otherwise interrupt announcement of the aria live content.

       - Improves the rapid typing situation in IE + JAWS, where fast user input can sometimes fail to trigger announcement.

       - Improves a situation in NVDA where stale results are sometimes announced. 
          e.g. when arriving at an initial result list and then filtering further by typing more characters, a situation can arise where the screen reader announces stale count information, e.g.
          '8 results are available, UK, 1 of 8 is selected' ..then ..
          'UK, 1 of 23'
          ..where the '23' relates to the size of initial result list.

- Introduced additional `aria-describedby` hint when input is empty (inspired by [haltersweb autocomplete](https://haltersweb.github.io/Accessibility/autocomplete.html), and recommended by DAC). The `aria-describedby` attribute is removed once user input is present to cut down on overly verbose screen reader announce.

- Silenced the aria live region when a valid choice has been made. (Where valid choice is detected as direct input of a valid value, active selection of a matched value from the result list, or the auto-select of currently highlighted value on focus blur when `autoselect` is enabled)

- Re-display menu on re-focus of input, whenever a matched query is present and no valid choice has yet been made. Aligns behaviour with the 'No search results' scenario, and also ensures navigation possibilities are aligned with the resulting aria live announcement.

- Introduced `aria-label` attribute on each list item element within the matched results (iOS only). The label includes the index of each item (e.g. 'France 2 of 7') to ensure this info is announced when exploring by touch/swipe. (By default iOS only announces 'list start' and 'list end' on first & last elements respectively.)


timpaul:
Thanks so much for this Mark! The team is focussed on releasing the next version of GOV.UK Frontend right now, but we'll be picking up this pull request for review after that - in about 2 weeks time.

Thanks again.

markhunter27:
No prob, it's been an interesting piece of work for us. Will check back for tech / functional feedback. I can make further refinements where required, or you can always fork my fork - all good :)

src/autocomplete.js
```diff
@@ -466,8 +466,10 @@ export default class Autocomplete extends Component {
         )}
 
         <input
+          aria-expanded={menuOpen ? 'true' : 'false'}
           aria-activedescendant={optionFocused ? `${id}__option--${focused}` : false}
           aria-owns={`${id}__listbox`}
+          aria-autocomplete='both'
```
36degrees: This change isn't mentioned in the commit message – does it have any perceivable effect? It seems like having moved the combobox role onto the text input we're following the [ARIA 1.0 pattern](https://www.w3.org/TR/wai-aria-1.0/roles#combobox) rather than [ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#combobox). In ARIA 1.1, authors MUST add `aria-autocomplete`, but this is a SHOULD in the 1.0 pattern. If this doesn't provide any tangible benefit, I'd consider leaving this out.

If we're going to keep it, from reading [the spec](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-autocomplete), it sounds like this should only be `both` if `autoselect` is `true` – that's the only time when the autocomplete 'presents a value completion prediction inside the text input'. Otherwise, it should be `list`.

src/status.js
```diff
@@ -56,11 +78,13 @@ export default class Status extends Component {
       content = tResults(length, contentSelectedOption)
     }
 
+    this.debounceContentUpdate()
```
36degrees: Is there a way to debounce the content update without the `render` function effectively (through a timeout) changing the component state and thus triggering another `render`?

We're wondering if there's a way to do it using the `componentWillUpdate` lifecycle event on the `Status` component?

Alternatively, avoid using the virtual DOM entirely, and update the status using the native DOM API?

src/status.js
```diff
@@ -20,6 +18,7 @@ var debounce = function (func, wait, immediate) {
 
 export default class Status extends Component {
   static defaultProps = {
+    statusDebounceMillis: 1400,
```
36degrees: Is this definitely something we think needs to be configurable? Is that something that we actually expect users to change, or just for testing purposes? Generally, it'd be good to try and avoid exposing props just for testing purposes as they then become part of the public API.

If we do think there's a good reason to keep this around, can we make sure it's added to the API docs?

src/autocomplete.js
```diff
@@ -520,6 +525,9 @@ export default class Autocomplete extends Component {
             <li className={`${optionClassName} ${optionClassName}--no-results`}>{tNoResults()}</li>
           )}
         </ul>
+
+        <span id='assistiveHint' style={{ display: 'none' }}>When autocomplete results are available use up and down arrows to review and enter to select.  Touch device users, explore by touch or with swipe gestures.</span>
```
36degrees: This will need to be translatable, like `tNoResults`.

src/autocomplete.js
```diff
@@ -57,7 +57,8 @@ export default class Autocomplete extends Component {
     required: false,
     tNoResults: () => 'No results found',
     dropdownArrow: DropdownArrowDown,
-    isQueryAnOption: (query, options) => options.map(entry => entry.toLowerCase()).indexOf(query.toLowerCase()) !== -1
+    isQueryAnOption: (query, options) => options.map(entry => entry.toLowerCase()).indexOf(query.toLowerCase()) !== -1,
+    isIosDevice: isIosDeviceDetected
```
36degrees: Again, it'd be good to avoid making this a prop just for tests - can we explore e.g. stubbing `navigator.userAgent` instead?

src/autocomplete.js
```diff
@@ -523,6 +526,9 @@ export default class Autocomplete extends Component {
             const showFocused = focused === -1 ? selected === index : focused === index
             const optionModifierFocused = showFocused && hovered === null ? ` ${optionClassName}--focused` : ''
             const optionModifierOdd = (index % 2) ? ` ${optionClassName}--odd` : ''
+            const ariaLabelProp = (this.props.isIosDevice()) ? {
+              'aria-label': `${option} ${index + 1} of ${options.length}`
```
36degrees: This makes me a little nervous, especially if the user is using a suggestion template, as what is being displayed to the user might not match `${option}`, and we're overriding it by providing a label.

I'm also wondering whether Voiceover supports `aria-posinset` / `aria-setsize` as suggested in #277 as this would remove the need to provide a label.

Alternatively, could we explore associating the position and number of results available using `aria-describedby` or any another method?

src/status.js
```diff
@@ -56,25 +81,51 @@ export default class Status extends Component {
...
+    this.debounceStatusUpdate()
+
+    return (
+      <div>
+        <div
+          id='ariaLiveA'
+          aria-atomic='true'
+          aria-live='polite'
+
+          style={{
```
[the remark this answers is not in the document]
markhunter27: Acknowledged - will tidy.

src/autocomplete.js
markhunter27: Good point, I'd missed this (consideration of the config option of `both` vs `list`).
I can either remove, or introduce the dynamic value based upon config.

I added the extra attribute during JAWS testing at one point. It didn't improve the situation I was focused on, but I didn't subsequently remove it on the basis of added semantics.
As far as I'm aware it isn't having any direct impact upon any assistive tech behaviour currently, but I'll double check this on my next round of testing (once I've actioned these comments)

Regarding the combobox role and ARIA 1.0 vs 1.1 - in this case it seemed necessary to strike a balance between latest spec, and what actually works with the assistive tech. I suspect the spec is ahead of what the tech is currently supporting. Moving the combobox role onto the input and moving away from the 'composite' approach was the change that brought the field to the attention of the form navigation tools (where it was ignored previously).  It's an awkward one for sure - spec compliance vs real world compatibility with screen readers.
Happy to discuss further.

src/status.js
markhunter27: I actually had a few variations around the debouncing - including using the `componentWillUpdate` lifecycle method at one point. I settled on the current form as the tidiest / most readable - but no issue with moving to the lifecycle if preferred.
Whichever approach, I need to avoid creating an infinite loop by ensuring the state is only amended one more time.

There's also the option to bring in [this](https://github.com/samuherek/use-debounce-preact) library, in preference to the local debounce function. I haven't investigated so far, as the local solution performs ok, but it may improve readability.

The debounce is bringing a couple of benefits in terms of JAWS and rapid typing, and in preventing the typing echo interruption in VoiceOver - so moving to direct DOM manipulation wouldn't itself resolve those issues - although it might avoid the need for the duplicate live regions.

src/status.js
markhunter27: Sure - will remove this from props.

src/autocomplete.js
markhunter27: Acknowledged - will revise.

src/autocomplete.js
markhunter27: Sure :)

src/autocomplete.js
markhunter27: I'll investigate the use of the ARIA attributes as an alternative.
Adam Liptrot (one of the other FE guys here) also made the same suggestion.

If that doesn't yield results, I can possibly use the same resolving function that is being used for the displayed value, to remove the possibility of discrepancy? Will look into.

src/autocomplete.js
markhunter27: Confirmed during tests that the `aria-autocomplete` is actually improving the screen reader announce in NVDA ..slightly (no difference in JAWS). It adds an extra 'has autocomplete' to the focus announce.
I've left the attribute in place but made the value dynamic based upon the `autoselect` configuration (and also iOS detection - as the existing util function already considers this).

src/status.js
markhunter27: I've experimented with the direct DOM update approach. Be good to discuss over hangout.
My changes for this one are parked on another branch - `markh-direct-dom`
It's the only change relating to these comments that I haven't applied directly to my master branch - pending discussion :)

src/status.js
markhunter27: Now removed from props.

src/autocomplete.js
markhunter27: Moved into translatable props, and README updated.

src/autocomplete.js
markhunter27: Changed as recommended. 
Have also moved the `isQueryAnOption` util function out of props (that one for a different reason although it equally didn't belong - now catering for the user defined suggestion template scenario, prompted by your other comment below..)

src/autocomplete.js
markhunter27: On testing, the `aria-posinset` and `aria-setsize` attributes unfortunately don't help the situation on iOS, ..but they **do** resolve the stale count problem in NVDA (as per the issue you've linked). So the extra attributes are now added.

In terms of the iOS touch/swipe I've settled on an alternative approach, conditionally appending a (visually hidden) `span` element to each `li` option element. The span contains the position info, e.g. `2 of 6`, and caters for the user defined template scenario by appending to the `dangerouslySetInnerHtml`  content - which is already populated from the `templateSuggestion()` resolver function.
Tested on iPhone - you can hit the 'custom results' instance on the test page on my github.io [here](https://markhunter27.github.io/accessible-autocomplete/examples/) to observe.

I did first try an `aria-describedby` approach. I rendered a second `ul` of description elements each referenced using the attribute, but saw some unusual behaviour - always on the first element of the list, where iOS and NVDA both announce `1 of 1` despite the content showing e.g. `1 of 3` in the chrome/FF inspector. It could be that an earlier render() invocation is setting the attribute before the options data is populated and the screen reader isn't budging once it has a value in it's hands. I tried a few different variations, and then moved to the suffix approach which felt tidier all round.

src/status.js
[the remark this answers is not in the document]
markhunter27: Now amended

README.md
```diff
@@ -316,6 +316,14 @@ Type: `Function`
 
 A function that receives two arguments, the count of available options and the return value of `tStatusSelectedOption`, and should return the text used in the accessibility hint to indicate which options are available and which is selected.
 
+#### `tAssistiveHint` (default: `() => 'When autocomplete results are available use up and down arrows to review and enter to select.  Touch device users, explore by touch or with swipe gestures.'`)
```
nickcolley: I've noticed in the other language we have stripped the end period, should we keep this new additional message consistent with this?

markhunter27: As discussed :) .. it seems to be a specific to the aria live region announcement that the ending period is announced as 'dot'
The hint here is read ok, and the full stop & comma introduce the expected pauses in speech.

PR #355, opened 2019-09-09
Refinements to address accessibility issues

markhunter27:
_**NOTE: This PR is a replacement for [this PR](https://github.com/alphagov/accessible-autocomplete/pull/348) which evolved through discussion & refinement. It applies the the same collection of amendments in a more structured set of commits.**_
&nbsp;
The changes are focused on improvements in the following browser / screen reader combinations:

- Internet Explorer and JAWS
- Firefox and NVDA
- Safari and VoiceOver
- iOS Safari and VoiceOver

..although we've also done some less focused testing using NVDA and JAWS readers with each combination of Firefox, IE, and Chrome.

The changes are in part refinements to the markup of the component (type and placement of ARIA attributes), and in part changes to the detail and timing of ARIA live region updates ('status' updates).

### Aria live region refinements to improve consistency of screen reader announcements

1. Now using a pair of aria live regions and flip-flopping status updates between. This change resolves the problem of duplicate screen reader announcements (many times, in rapid succession), that are caused by undesirable React virtual DOM behaviour. See [here](https://github.com/nvaccess/nvda/issues/7996) for background inspiration. The change also serves as a refinement to the previous 'comma' approach (see [here](https://github.com/alphagov/accessible-autocomplete/issues/200#issuecomment-316755847) for original discussion), in that it also addresses the problem of ensuring a new status announcement on every user input (even when where result set has not changed with the new input). 

2. Introduced debouncing of status updates (aria live region updates). This delays and limits frequency of screen reader announcements, and has the following benefits:
   - Improves situation on Mac Safari + VoiceOver, where 'typing echo' can otherwise interrupt & mask announcement of the aria live content.
   - Improves the rapid typing situation in IE + JAWS, where fast user input can sometimes fail to trigger a screen reader announcement.
&nbsp;
3. Small refinements to content within the aria live regions:
   - Removed full stops, which are announced as 'dot' by VoiceOver on Mac
   - Follwed DAC language recommendations to change 'selected' to 'highlighted'
&nbsp;
4. Removed inner `span` element within each aria live region. This small change resolves a problem in JAWS + IE, where the screen reader wouldn't otherwise announce the live region unless the user first disables "virtual PC cursor" mode (which is on by default).

### Markup refinements to the HTML `input` element to improve form behaviour

1. Moved the `combobox` role to the `input` element and away from the wrapping `div`. This is a regression to the [ARIA 1.0](https://www.w3.org/TR/wai-aria-1.0/roles#combobox) pattern rather than [1.1](https://www.w3.org/TR/wai-aria-1.1/#combobox), but screen reader support found to be much better  in testing. (The `input` form control becomes visible to, and navigable by the screen reader with this change - where it was invisible before. E.g. when using VoiceOver rotor, or NVDA Insert + F7 form controls mode.)

2. In a similar motivation, moved the `aria-expanded` attribute onto input element rather than the composing `div`, and added an `aria-autocomplete` attribute. Both of these changes bring about improvements in NVDA, where the expanded/collapsed status is now announced, together with a 'has autocomplete' phrase when the input receives focus.

3. Introduction of an additional `aria-describedby` 'hint' when the input is empty - inspired by [haltersweb autocomplete](https://haltersweb.github.io/Accessibility/autocomplete.html) component and recommended by DAC. The `aria-describedby` attribute is removed once any user input is present, to cut down on screen reader verbosity.

### Refinements to component behaviour when a valid choice has been made

1. Introduced silencing of the aria live regions once a valid choice is detected in the input. This helps to prevent potentially confusing screen reader noise once the user has supplied a suitable value.

2. Amendment to ensure display of the options list on focusing of the input, conditional upon a matched query being present in the input, but no valid choice having yet been made. The change improves consistency between the audio announcements and navigation possibilities.

### Refinements to the options list to improve count and index announcements

1. Addition of `aria-posinset` & `aria-setsize` attributes to each `li` suggestion element. This change resolves a problem in NVDA where a stale result count can sometimes occur as the user continues to type and constrain the suggestion list (see [this issue](https://github.com/alphagov/accessible-autocomplete/issues/277)). It also improves announcement slightly in JAWS + IE, where the index phrase (e.g. '2 of 12') now appears earlier in the screen reader statement (albeit duplicated!)

2. Addition of a visually hidden `span` element to each `li` content - specifically on iOS. The added suffix contains index info (e.g. '2 of 12'), to improve touch/swipe announcements which don't otherwise include position (other than 'list start' and 'list end').
&nbsp;
### Screen reader audio recordings

NVDA version 2019.2 on windows 10, with Firefox version 69.0
JAWS version 2019.1907.42 ILM on windows 10, with Internet Explorer version 11.1.18362.0 
VoiceOver on Mac Mojave version 10.14.5, with Safari version 12.1.1
VoiceOver on iOS version 12.0, with Safari mobile browser

[screenreader_recordings_post_amend.zip](https://github.com/alphagov/accessible-autocomplete/files/3588432/screenreader_recordings_post_amend.zip)

### github.io 
Amended component is [deployed here](https://markhunter27.github.io/accessible-autocomplete/examples/form-single.html)


src/status.js
```diff
@@ -56,25 +56,35 @@ export default class Status extends Component {
...
+          aria-atomic='true'
+          aria-live='polite'>
+          <span>{bump ? content : ''}</span>
+        </div>
+        <div
+          id='ariaLiveB'
+          role='status'
+          aria-atomic='true'
+          aria-live='polite'>
+          <spam>{!bump ? content : ''}</spam>
```
nickcolley: Is it possible to fix this typo?

markhunter27: Dammit! .. ha ha, how the hell did I miss that.

markhunter27: Ok done.

markhunter27: Hey Nick, cheers for checking over the commits, appreciated.

Happy to help with issue categorisation where I can.  May have to step back from the detail for a while, but feel free to ping me with plan of attack. :thumbsup:

Issue #359, opened 2019-09-11
Update focus styles to match GOV.UK Frontend 3.0.0


nickcolley:
## What
The current focus style and colours are out of date when compared with GOV.UK Frontend 3.0.0, which means they do not meet the new WCAG 2.1 non-text requirements.

## Further detail
We should make sure the styles match the guidance on focus state: https://design-system.service.gov.uk/get-started/focus-states/

There is a related discussion here: https://github.com/alphagov/accessible-autocomplete/pull/329 but we may decide to do something more pragmatic in the short term.

## Done when

- [ ] Focus styles match GOV.UK Frontend inputs

PR #329, opened 2019-02-18
WIP: Discussion: Separate default and GOV.UK styles

dashouse:
I was going to do a PR to update the input style from **GOV.UK Elements** to **GOV.UK Frontend** but when looking at the multiple uses of this component form an open source perspective I thought it could be useful to break the styles up into "Default", a version of the autocomplete with basic brand-free styling with separate overrides for "GOV.UK Frontend" styles.

There may be a better way to achieve this technically, but I've done the PR as a proof of concept / idea.

### Default / generic styles
screen shot 2019-02-18 at 08 52 49
screen shot 2019-02-18 at 08 52 56

### GOV.UK overrides (Updated to GOV.UK Frontend size / padding / responsive text)
screen shot 2019-02-18 at 08 54 28
screen shot 2019-02-18 at 08 54 34

PR #360, opened 2019-09-11
Update focus styles to meet WCAG 2.1 level AA non-text contrast requirements

nickcolley:
This is a modest attempt at updating the autocomplete to address the WCAG 2.1 non-text related issues around the current focus by making it align with GOV.UK Frontend version 3.0.0.

Makes the component's focus styles contrast [without relying on colour alone which means the new non-text contrast guideline in WCAG 2.1 level AA](https://design-system.service.gov.uk/get-started/focus-states/).

Also [updates other colours from the GOV.UK palette to the new colours](https://designnotes.blog.gov.uk/2019/07/29/weve-updated-the-gov-uk-colours-and-font/).

## Screenshots
| Before | After |
| - | - |
|  |  |
|  |  |
|  |  |

There are some more interesting ideas about the future of theming for this component here: https://github.com/alphagov/accessible-autocomplete/pull/329 but this aims to be a small change so we can release it soon.

Fixes https://github.com/alphagov/accessible-autocomplete/issues/359


nickcolley:
Labelled this as a breaking change as I think the styling could catch people out who are importing this directly from npm

src/autocomplete.css
```diff
@@ -20,17 +20,17 @@
...
 .autocomplete__hint {
-  color: #BFC1C3;
+  color: #b1b4b6;
   position: absolute;
 }
 
-.autocomplete__input--default{
-  padding: 4px;
+.autocomplete__input--default {
+  padding: 5px;
```
36degrees: Should the autocomplete have a `height: 40px` to match our text inputs? At the minute it's 39px which is frustratingly close.

nickcolley: I've tried not to change that since this component relies on layering of inputs and overlays for some features and it felt risky to change, so wanted to do a simple update and maybe do something better if we decide to do the stuff Dave was proposing in the other PR.

tvararu: I'd do it :shipit:

CHANGELOG.md
```diff
@@ -2,6 +2,30 @@
...
+
+You can [copy the CSS from the dist folder](https://github.com/alphagov/accessible-autocomplete/blob/master/dist/accessible-autocomplete.min.css).
+
+#### If you're using Sass imports
+
+This will be updated automatically
+
+#### If you have custom additional CSS
+
+You should check that your custom CSS does not remove the improvements to the focus state, if you do need to change the appearence keep in mind the WCAG 2.1 level AA requirements.
```
tvararu:
```suggestion
You should check that your custom CSS does not remove the improvements to the focus state, if you do need to change the appearance keep in mind the WCAG 2.1 level AA requirements.
```

src/autocomplete.css
nickcolley: This pull request isn't meant to be making the component 'proper GOV.UK Frontend' so would be nice to get this in and improve it later in the way Dave has suggested, will see what Ollie thinks 😬

PR #385, opened 2019-10-31
 Set aria-activedescendant to automatically selected option when using autoselect

peterjaric:
When using the arrow keys to navigate, aria-activedescendant is set to the currently focused option. But when "autoselect" is set to true and the first option is automatically selected because of a matching query string, it is not focused and as a consequence aria-activedescendant is not set. This PR tries to address that.

I have not added a test for this change since Karma is totally new for me and when running the tests not even the existing autoselect tests are run (in my environment, maybe I'm just using the wrong command). So, since the change is quite small, and I don't have time to learn Karma before adding a test, I have skipped that.


peterjaric:
Ah, some more commits were automatically added to the PR. Sorry about that, they were meant to go into a separate PR. Will research if I can remove them from this one.

nickcolley:
Hello @peterjaric,

Thanks for raising this, it's an interesting one.

Do you know what impact this has on people using a screen reader?

We're concerned that by setting this without the user moving to the listbox that it may cause confusing announcements implying that the focus has moved when it has not.

peterjaric:
It turns out we're not going to use the autoselect feature after all, and because of that I don't think I can spend more time working on this particular code. So you may close the PR, or keep it if someone else wants to take a stab at testing around a little.

nickcolley:
That's OK, everything is recorded even if we close things so it'll be helpful for us in the future, no wasted time :)

Thanks for helping us out.

PR #386, opened 2019-10-31
Add aria-controls

peterjaric:
According to https://www.w3.org/TR/wai-aria-1.1/#combobox "when the combobox is expanded, authors MUST set aria-controls on the textbox element to a value that refers to the combobox popup element". This change was prompted by an error message when running the SiteImprove Accessibility Checker extension.

I have added aria-controls to the input element. Currently aria-owns is used, but when I read about the difference between these two attributes I think that aria-owns probably is meant to be used in a different scenario (source: https://stackoverflow.com/a/34015524/185596). I have left aria-owns as it is, since there may be a very good reason that I am unaware of why it's there, but I can remove it in this PR if needed.


nickcolley:
Thanks for proposing this improvement.

In the version 2.0.0 release (https://github.com/alphagov/accessible-autocomplete/pull/355) we have moved from an implementation based on the ARIA 1.1 recommendations to ARIA 1.0, this is because there is better support from assistive technologies with the older pattern.

In the 1.0 specification we use `aria-owns` for similar functionality that the 1.1 specification uses `aria-controls`.

You can see details on how the ARIA 1.0 specification works here:
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.0pattern/combobox-autocomplete-both.html

If you find that adding this attribute noticeably improves behaviour in real world usage then please let us know more and we reconsider this.

peterjaric:
I see, thanks for the reference to the 1.0 spec, and thanks for the very informative background info. I'll also mention this to my team!

PR #388, opened 2019-11-04
Aria selected as string

peterjaric:
The aria-selected attribute of the `li` element is dropped (at least in Chrome 78.0.3904.70 on Windows 10). My guess is that it is Preact that does this with boolean attributes. 

I have changed it to use as string instead ("true" or "false"), in line with the aria-expanded attribute in the `input` element.


nickcolley:
Hey @peterjaric,

Thanks again for these issues.

This does seem inconsistent with the other use of boolean aria attributes such as `aria-expanded`.

We're happy with your suggestion but ideally it would be good to have test coverage for this change so we do not accidentally break this in the future.

Is this something you'd be able to do? When we have time we could help with this if you are not able to.

Do you know if this change improves the component when using a screen reader?

peterjaric:
Hi @nickcolley,

Great feedback again!

I will try to get the tests running (when I run them, not the same tests are run as in the CI, so I must do something wrong) and then I'll add a test for this functionality.

I haven't checked the effect on a screen reader (as you may have guessed, I'm trying to satisfy a set of automatic accessibility checks, rather than actually using a screen reader. There is some potential for improvement in my process for sure 😄 ), but I will try to do that too.

peterjaric:
May I suggest that you add a line about testing in a screen reader before creating the PR, i CONTRIBUTING.md? Under "PR nice to haves" perhaps.

peterjaric:
I have now added a test. I'm not sure it went into the right place, but at least it fails for the old code and succeeds for the new code. (While writing this comment, I see that the tests fail on Travis, so I guess I have some more work to do. But I'll finish my remarks here first.)

I also tested both versions with NVDA on Windows, and as far as I could tell there was no difference (I listened but also used the "Speech viewer" to be sure).

Some off-topic remarks about running the tests that might be good to know:

My dev environment is WSL, i.e. Ubuntu running in the Windows Subsystem for Linux, and after having had uncountable problems with all sorts of things, but finally getting almost everything to run, I switched to npm and node in Windows instead and everything worked (almost). So if someone else reports these kind of problems, ask if they are runing under WSL.

What didn't work was two things - first, if you run a newer version of npm and node, fibers (which is a dependency) doesn't build, so you have to go into package-lock.json and upgrade it to at least 4.0.0 (or even 4.0.2). See https://github.com/laverdet/node-fibers/issues/409 for some more details.

Second - puppeteer isn't automatically installed, you need to install it separately. Maybe this is intended, but anyway, when you do, the fibers version is reset in package-lock.json and you need to fix it again. Both this problems were present in WSL too, btw.

nickcolley:
In terms of the Windows specific issues, you could consider raising those as a separate issue to improve running this project in Windows.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Fixes
 
 - [Pull request #379: Ensure multiple autocompletes on one page do not have conflicting id attributes](https://github.com/alphagov/accessible-autocomplete/pull/379)
+- [Pull request #388: Set aria-selected as a string instead of a boolean to avoid being dropped.](https://github.com/alphagov/accessible-autocomplete/pull/388)
```
nickcolley: Sorry, one thing I missed, could you move this CHANGELOG entry to the 'unreleased' section under a new 'fixes' heading please? Then I'll get this merged :)

peterjaric: Fixed (it's obvious now that you mentioned it)! Sorry for the messy history, though...

nickcolley:
Closing this as we've merged the commits in another pull request, thanks for your help on this. :)

Issue #394, opened 2019-11-18
Saucelab tests are failing on the master branch


36degrees:
The Saucelab tests are currently failing on the master branch:

```
element click intercepted: Element <strong>...</strong> is not clickable at point (263, 362).
Other element would receive the click: <li aria-selected="false" class="autocomplete__option" id="autocomplete-customTemplates__option--0" role="option" tabindex="-1" aria-posinset="1" aria-setsize="1">...</li>
(Session info: chrome=78.0.3904.70)\n (Driver info: chromedriver=78.0.3904.70 (edb9c9f3de0247fd912a77b7f6cae7447f6d3ad5-refs/branch-heads/3904@{#800}),platform=Windows NT 10.0.10586 x86_64)
```

They appear to have been failing since #393 was merged.

peterjaric:
Sorry about this, I overlooked the Sauce Labs section in CONTRIBUTING.md.

36degrees:
@peterjaric please don't worry about it – it's a known issue that because SauceLabs requires an API key, the tests don't run on contributions from external contributors. Really we need to improve our processes to try and catch this stuff earlier, but in the meantime we expect this to come up occasionally.

Thank you again for your contribution – we really appreciate you helping to improve the accessible autocomplete 🙂

peterjaric:
OK, I'll let you handle it then! 

(Also since I got this error message after creating a free trial account at Sauce Labs and running the tests: "ERROR: Uh oh, you've run out of minutes!" 😄 )

nickcolley:
Just investigating this, I don't think that pull request was the cause of these tests failing, it seems to be unrelated.

nickcolley:
I was curious if we're just seeing general flakeyness in the tests so I triggered a rebuild of master tests.

```
1) Accessible Autocomplete basic example should announce status changes using two alternately updated aria live regions:
expected the first aria live region to be populated within 10000 milliseconds
running chrome on Windows 10
Error: expected the first aria live region to be populated within 10000 milliseconds
    at elementIdText("0.5907962755285172-4") - getText.js:35:50
```

Which failed on a known flakey test. Going to try once more and see what happens.

nickcolley:
Hmm, this time Travis completes and it looks like everything is fine but the Saucelabs dashboard reports failures? 🤔

nickcolley:
OK, so I think it is a problem introduced by that pull request, the tests rely on a specific ordering currently so the new test broke this assumption.

[...]

PR #396, opened 2019-12-02
Only run flakey test in Chrome

nickcolley:
I've tried a bunch of things to try and make these tests run consistently and have decided the only run this test in Chrome as I don't see the value in running it cross browser. Ideally the test suite would not be flakey so we dont have to do this. This is up for discussion on if this is the right thing to do.

Fixes https://github.com/alphagov/accessible-autocomplete/issues/394

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -24,7 +24,7 @@ const basicExample = () => {
 
     it('should show the input', () => {
       browser.waitForExist(input)
-      expect(browser.isVisible(input)).to.equal(true)
+      expect(browser.isVisible(input)).to.equal(false)
     })
 
     it('should allow focusing the input', () => {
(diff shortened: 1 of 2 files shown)


nickcolley:
It seems like the webdriver command to refresh is not consistently working, the command is sent to saucelabs but the browser does not seem to be refreshing.

nickcolley:
I'm confused why the dashboard for Saucelabs says there are failures but Travis says all tests pass, maybe it's expected that Saucelab will report failures as long as it returns the correct results to the test suite?

nickcolley:
I've somewhat admitted defeat and reduced the scope of the tests...

nickcolley:
Hmm, so I've really been round in circles on this one but my current thinking is that the Saucelabs dashboard can show failures for flakeyness but the test suite itself only cares about what is returned.

I've tested intentionally failing a test to confirm that Travis fails when something is not right, so this leads me to thinking the Saucelabs dashboard is not useful to look at but instead we should just focus on if the Travis build returns successful.

PR #397, opened 2019-12-09
Remove Saucelabs build dashboard

nickcolley:
We expect failures from flakey tests and intentionally failing tests so this gives a false representation of the status of the build.

The tests run in Travis do not care about if Saucelabs is flakey, only that the response is as expected.

Closes https://github.com/alphagov/accessible-autocomplete/issues/394


36degrees:
Based on https://github.com/alphagov/accessible-autocomplete/issues/394#issuecomment-570256605 I'm not sure we want to do this – I think the failing tests in SauceLabs are 'legit', and there's something else going on with the way they're being reported in Travis.

nickcolley:
Closing for now

Issue #398, opened 2019-12-16
Accessible autocomplete requires style-src 'unsafe-inline' and 'unsafe-eval' to work on pages protected by Content Security Policy (CSP)


alexbishop1:
The accessible autocomplete does not seem to play very nicely with [Content Security Policy (CSP)](https://www.w3.org/TR/CSP2/), specifically the [`style-src`](https://www.w3.org/TR/CSP2/#style_src) directive:

1. On iOS browsers, it sets `style` attributes on elements, which is blocked by the `style-src` directive unless `'unsafe-inline'` is specified
2. It uses the `cssText`setter as part of a function to detect support for pointer events, which is blocked by the `style-src` directive unless `'unsafe-eval'` is specified (curiously, only some browsers seem to report CSP violations for this)

Therefore, using the accessible autocomplete on pages protected by CSP only works if one allows `'unsafe-inline'` and `'unsafe-eval'` as style sources, which reduces the protection offered by CSP.

This was discovered when trying to apply CSP to GOV.UK Pay’s enter card details page, which uses the [GOV.UK country and territory autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete), which is built on the accessible autocomplete component.

nickcolley:
Looks like there may be other ways to detect pointer events: https://github.com/rafrex/detect-pointer-events/blob/master/src/index.js

https://github.com/alphagov/accessible-autocomplete/blob/fe3ea113a623bc0b21c7dd2a698f3d315bf5b747/src/autocomplete.js#L16-L22

I'm surprised that code trips unsafe-eval though!

nickcolley:
Hmm on second thoughts they're detecting related features in 

- JavaScript: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/pointerevents.js
- CSS: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/pointerevents.js

nickcolley:
Looks like we may be able to use a different method for example `Object.assign` like so:

https://github.com/Popmotion/popmotion/pull/659/files#diff-69eb3a42a2b4f328f0e78dc96a6fc302R35

From https://github.com/Popmotion/popmotion/issues/592

alexbishop1:
> I'm surprised that code trips unsafe-eval though!

https://www.w3.org/TR/CSP2/#style_src calls out `cssText` as being blocked by `'unsafe-inline'`.

nickcolley:
[...]

PR #400, opened 2019-12-16
Remove pointer events check

nickcolley:
It's not clear to me why this functionality was added since the component does not use pointer events on the hint (https://github.com/alphagov/accessible-autocomplete/blob/master/src/autocomplete.css) so I'm attempting to remove it, and see if there's another way to resolve issues with IE9 that don't break CSP

> Don't display <Hint /> on browsers without pointer events (fixes IE9 tests)
https://github.com/alphagov/accessible-autocomplete/pull/101

I've tested the autoselect example in IE9 and it works as expected.

Related: https://github.com/alphagov/accessible-autocomplete/issues/398


nickcolley:
Theo:

> yeap, if I recall correctly, IE9 does not overlap inputs nicely
> 
> the hint text is based on an overlayed input as you know (edited) 
> and it was misbehaving in IE9

Will do some more testing in IE9 and see if I can reproduce this.

nickcolley:
Had another look at this, I think this is safe to remove for users of IE9 but it may be failing a Saucelabs test, worth checking once we've resolved the flakey saucelabs dashboard.

hannalaakso:
This makes sense to me as since @nickcolley pointed out, it seems that the pointer events check was added to resolve failing tests, rather than fix IE9 itself (https://github.com/alphagov/accessible-autocomplete/pull/101#issue-119168009). So if the tests pass then it looks 👍

nickcolley:
Will rebase and see if the tests pass.

nickcolley:
@hannako needs an official approval :)

hannako:
@hannako is me Jessica Jones, @hannalaakso is Hanna! Sorry for my weird username!!

nickcolley:
Damnit sorry @hannako :(

PR #401, opened 2020-01-02
Dismiss open autocompletes between tests

36degrees:
Clearing the value doesn’t seem to work reliably when the autocomplete is currently open, especially if it has been interacted with (e.g. in a test that sends the arrow down key to focus on a result)

By sending ‘escape’ we dismiss any open autocompletes, which then seems to make the input-clearing work reliably.


nickcolley:
Going to merge this so I can investigate https://github.com/alphagov/accessible-autocomplete/pull/400, hope this is OK :)

PR #403, opened 2020-01-14
Remove retries for functional tests

36degrees:
The [wdio-sauce-service](https://github.com/webdriverio/webdriverio/tree/83dbe5d5c12fdb91b12db8c02cf2b8b50cfbd608/packages/wdio-sauce-service), which is responsible for reporting the build state back to Saucelabs, does not seem to take into account retries – we _believe_ that if a test fails, it will mark the build as failed in Saucelabs even if it is then retried and passes. This is confusing, as it does not match what we see in the Mocha reporter either locally or in the Travis build logs.

Using retries also masks situations where for example the autocomplete state is not reset correctly between tests, and thus the first attempt of a test always fails.

For these reasons, we're going to try removing retries and see what problems it causes (or, to look at another way, unmasks)

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -1,4 +1,4 @@
-/* global afterEach, before, beforeEach, browser, describe, it */
+/* global afterEach, beforeEach, browser, describe, it */
 const expect = require('chai').expect
 const { browserName, version } = browser.desiredCapabilities
 const isChrome = browserName === 'chrome'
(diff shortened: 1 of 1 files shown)


nickcolley:
So that master Travis isnt failing should we solve the issues before merging this?

I'm in favour of removing retries though given the confusion.

test/integration/index.js
```diff
@@ -47,29 +37,51 @@ const basicExample = () => {
...
+          )
+
+          browser.addValue(input, 'h')
+          browser.waitUntil(() => { return (regionA.getText() !== '' && regionB.getText() === '') },
+            liveRegionWaitTimeMillis,
+            'expected the first aria live region to be populated, and the second to be cleared within ' +
+            liveRegionWaitTimeMillis + ' milliseconds'
+          )
+        } else {
+          browser.addValue(input, 's')
```
nickcolley: You could consider storing the result of `regionA.getText()` above and reducing the duplication here but don't think it's that important. It's arguably easier to understand what's going on as is.

PR #406, opened 2020-01-17
Make hint padding match input padding

### LINKED ISSUE #405: Misaligned hint content when showAllValues used
It looks like there's a bug when showAllValues is used. In that case, the input and the hint are off by 1px causing the text to look blurry / bold.

Example:
Screenshot 2020-01-16 at 17 44 52

It looks like this issue is this line, which sets a padding of 4px when the regular input has a padding of 5px:
https://github.com/alphagov/accessible-autocomplete/blob/4ad21bbe376b9b9507c6ab1c49097d8b6af2da52/src/autocomplete.css#L37

Changing it back to 5px fixes the issue for us.

edwardhorsford:
Fixes #405 

I reckon this should be a variable used by both rather than magic numbers, but it's css so ¯\_(ツ)_/¯


nickcolley:
Nice catch this makes sense!

Could you please add a CHANGELOG entry for this? If you prefer I can do this for you :)

edwardhorsford:
@nickcolley I've updated the changelog and added a comment to the line so future authors know it should match the other input.

nickcolley:
Nice work 👍

36degrees:
@edwardhorsford are you able to rebase this against master? That should then get the tests passing so we can get this merged 🤞

edwardhorsford:
@36degrees should be good to go.

36degrees:
Rebased to remove merge commit, update comment style so that the CSS can be compiled, and update dist.

edwardhorsford:
Thanks @36degrees!

PR #407, opened 2020-01-20
Use div for wrapper

### LINKED ISSUE #402: Question: why wrap in a span element?
Hello

Thank you for this code. Clearly a lot effort has gone into it.

Hope it's ok to ask here (I have read as much documentation as I can find) but is there any particular accessibility or other reason why, at least when using the `.enhanceSelectElement()` wrapper, the `autocomplete__wrapper` `<div>` itself is wrapped in a `<span>` element?

Regards

nickcolley:
Div elements are more appropriate to use for 'flow content' since span elements expect 'phrasing content' or text related elements.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element

Closes #402


36degrees:
Do we think this definitely closes #402? It definitely seems like an improvement to use a `<div>`, but you could still reasonably ask why there's any wrapper element at all…

nickcolley:
@36degrees the function expects the select element, and then puts a new element alongside the element as a place for React to mount onto. So it is required as you can't nest the autocomplete markup inside the original select.

You could consider changing the component so that it initialises an element that expects a selected nested within it but React likes to be able to control the entire DOM tree where it's mounted so it likely wouldnt work either.

So the TL;DR: Yes this is needed, but shouldnt be a span.

inspector71:
> 
> 
> @36degrees the `enhanceSelectElement` function expects the select element, and then puts a new element alongside the element as a place for React to mount onto. So it is required as you can't nest the autocomplete markup inside the original select.
> 
> You could consider changing the component so that it initialises an element that expects a select element nested within it but React likes to be able to control the entire DOM tree where it's mounted so it likely wouldnt work either.
> 
> So the TL;DR: Yes this is needed, but shouldnt be a span.

Is React compliance a design goal? I have seen references to React loosely here or there but am not completely certain if this component is primarily targeted at React. As I am not using React, I might need to find a different widget if the development is first and foremost aimed at React integration.

inspector71:
> > @36degrees the `enhanceSelectElement` function expects the select element, and then puts a new element alongside the element as a place for React to mount onto. So it is required as you can't nest the autocomplete markup inside the original select.
> > You could consider changing the component so that it initialises an element that expects a select element nested within it but React likes to be able to control the entire DOM tree where it's mounted so it likely wouldn't work either.
> > So the TL;DR: Yes this is needed, but shouldn't be a span.

Curious if the `div` would ideally be needed in that it wraps a "wrapper".

Issue #408, opened 2020-01-20
Issue with selecting items from dropdown


martinlex:
We have a kind of special situation:
My colleague can fail to select an item from the list, whereas I cannot.
When she clicks, it behaves like pressing "ESC", meaning: the item is not selected, it just keeps whatever you wrote, and closes the dropdown.
I, on the other hand, cannot recreate this bug. We have tried with Mac (both clickpad and mouse) and Dell laptop (both clickpad and mouse).
It seems to me, when I click very gently and very softly, I can recreate it (sometimes).
It sounds so silly, I know 😂😂
Is there a known problem with "light/short clicking" and "hard/long clicking"??? 
Don't know where to go from here, please help.

nickcolley:
I think this might be related to : https://github.com/alphagov/accessible-autocomplete/issues/328

martinlex:
Oh, thanks! So we were not alone! It looks like the pull request for this fix has merge conflicts. Will this fix be approved and merged later on?

nickcolley:
@martinlex I've asked the team about this, we didnt plan on fixing this bug as we don't have much time right now.

That pull request may or may not fix this issue so we'd need to do more work.

If I don't get back to you sooner I've put 'awaiting-triage' on this card which we use to keep track of cards we need to talk about as a team.

Thanks for your patience.

martinlex:
@nickcolley 
Thanks Nick. Meanwhile, is there a fix I can apply locally? We are loading from npm module.

nickcolley:
@martinlex I'm not aware of a quick easy fix since I don't understand why this problem is happening, sorry I can't be more helpful sorry Martin!

martinlex:
I understand. The PR #342 adds a mouseDown event
`onMouseDown={(event) => this.handleOptionClick(event, index)}`
Have you tried to see if that is the solution?
Thank you for the feedback 👍

nickcolley:
I'll have a poke around to see what I can learn and let you know if there's anything helpful short term/ long term.

PR #409, opened 2020-01-20
Fix long click issue

nickcolley:
This is an exploration to try to solve the issue with clicks failing on options: https://github.com/alphagov/accessible-autocomplete/issues/408

I believe that this issue was introduced when we added timers to this area of the code, which makes sense to me that this is why the length of a click matters:
https://github.com/alphagov/accessible-autocomplete/commit/a404a444b7775eb9ca93a19ed6f31d26aa1648a6

I've also referenced the other pull request that tries to solve this:

https://github.com/alphagov/accessible-autocomplete/pull/342/files

Will need to clean up the commits and comment something better I think if this is what we go with.

Approach is to use the relatedTarget on blur events which references the element that's being focused, then look through all the elements in the autocomplete and keep the focus inside the component if the new thing being focused isnt outside the component.


nickcolley:
I've also noticed that I can't reproduce the change that originally removed the mouseevent prevent, so going to explore that more because if that's the case then we can revert to the earlier known working version... 🤞 

https://github.com/alphagov/accessible-autocomplete/pull/310

PR #415, opened 2020-02-06
Make React bundle work server-side

### LINKED ISSUE #352: Server side rendering react component fails
Server rendering of the react component fails in two places, because it's looking for either `window` or `document` (I'll point out where in the code, below).

This would happen with either a universal (isomorphic?) app or statically pre-rendering react components into HTML, for example with Gatsby.

### Repro steps
1. Create a Gatsby application.
2. `import AccessibleAutocomplete from "accessible-autocomplete/react";`
3.  Render `<AccessibleAutocomplete ...`

### Expected result
The render works OK, rendering static markup on the server that is rehydrated on the client

### Actual result
The following errors occur:
1. _window is not defined_ because of a webpack issue
2 _document is not defined_ because of the `hasPointerEvents` function

## Reason for window being undefined
This is because of issues with webpack 4 in the way that the UMD bundle is generated:
webpack/webpack#6642
webpack/webpack#6677
The fix can be as simple as: https://github.com/ankurk91/vue-flatpickr-component/commit/c6f37baec05c6c58ef4f7ec0b13cea73aeedfeb2

## Reason for document being undefined
The `hasPointerEvents` function is called when the module loads, and looks for `document.createElement`
See https://github.com/alphagov/accessible-autocomplete/blob/master/src/autocomplete.js#L18

andymsuk:
# Overview
Fixes #352, so that the component can be used in a server-side React environment. 

# Changes
- Set `globalObject` to `'this'` in the React Webpack config, so that it doesn't have unwanted references to `window`, in order for it to work in a server-side environment
- Check `navigator` is defined before using it, in order for it to work in a server-side environment
- ~~Only include the `role` of `combobox` once the component has mounted, so that the server-side rendered markup (i.e. when the user has JavaScript turned off) has no `role` attribute, and so it's role defaults back to `textbox`~~
- ~~Only include the aria hint (`aria-describedby`) once the component has mounted, so that the server-side rendered markup (i.e. when the user has JavaScript turned off) has no `aria-describedby` as it doesn't make sense if there the JavaScript is not running.~~


nickcolley:
Thanks a lot for helping with this.

One thing that worries me is this change:

> Only include the role of combobox once the component has mounted, so that the server-side rendered markup (i.e. when the user has JavaScript turned off) has no role attribute, and so it's role defaults back to textbox

As we discussed in the issue, we probably want a completely different component to be rendered server side. We want people to send the markup from the [progressive enhancement](https://github.com/alphagov/accessible-autocomplete#progressive-enhancement) section to users before loading the JavaScript, instead of sending a variation of the enhanced component.

What're your thoughts on ensuring that we don't encourage folks rendering a broken experience server side?

nickcolley:
If you look at their example will give more context:

https://github.com/nice-digital/global-nav/blob/master/src/Header/Search/Autocomplete/Autocomplete.jsx#L133

nickcolley:
Perhaps some guidance in the README explaining server rendering could help.

ediblecode:
Thanks for looking at this. Must apologize though, been meaning to look at this for aaaages after I raised the issue! Good timing though as I'm picking the work up again next week.

For what it's worth I agree with the comment about "we probably want a completely different component to be rendered server side" so I think this fix should probably just be all the bits that break loading this component in a Node environment.

andymsuk:
Hi @nickcolley and @ediblecode 

Thanks for your comments.

With regards to wanting a different component rendered server-side, I have a couple of thoughts. 

My first thought is that it really depends on your use case for this component as to what you'd want rendered server-side for the case where JS is turned off or not loading for whatever reason. For example, in my personal use case, I want to use this component on my app merely as a way to suggest values to the user, so that they don't have to type the full value every time. In this case, just returning a default `textbox` type of input would be suitable. However, I appreciate that another use case is that the user is obliged to pick a value from the component, in which case I'd agree that a `select` is a more suitable fallback. 

My second thought is that if you're rendering a `select` server-side and then switching it out client-side for an `input type="text"` then, unless you've got your CSS spot on (and even still possibly in some browsers) the user will see a "flash" as the component changes. This "flash" effect may not be an option for some people using this component.

As you both suggested, I think the server-side fallback could do with some more thought, possibly having multiple options to account for the use cases I outlined above. I've therefore removed the last commit around changing the server-side fallback, so that this can be addressed either directly by the person using the component in their app, or by us or anyone else in a future enhancement PR. At least this way, as @ediblecode says, this fix can just be the bits that break loading this component in a Node environment.

nickcolley:
@andymsuk thanks for updating that, it's looking good I think the only last things to do are:

- tidy up the commits
- add a CHANGELOG entry

I can do this on your behalf if you want, let me know.

andymsuk:
Thanks @nickcolley.

I've tidied up the commits and added to the CHANGELOG, but feel free to edit/tidy more if you'd like.

ediblecode:
Looks good to me, have approved.

PR #449, opened 2020-06-19
Use build script for integration testing

lfdebrux:
This PR changes the integration test pipeline to generate assets
using the production build script before testing.

Why is this needed?
-------------------

`wdio-webpack-dev-server-service` has not been updated since 2018, and does not work with WebdriverIO v5 and up. The use of this service is blocking upgrading WebdriverIO.

I think this commit also fixes a bug in the integration test pipeline where changes to the source files would not necessarily be reflected in the system under test. While trying to upgrade the WebdriverIO tests in this repo I found that the `wdio-webpack-dev-server-service` was not working as expected; if assets were removed from the `dist` folder before running the tests the tests would fail as `wdio-webpack-dev-server-service` would not replace them.

What does this commit change?
-----------------------------

This PR uses the existing webpack pipeline to generate assets, rather than use the `wdio-webpack-dev-server-service`. I think this is better overall however. As well as changing the npm scripts in a hopefully backwards-compatible way, the WebdriverIO config has been updated to remove the use of the service.

As the assets in the dist folder were already being used by the example HTML used in the integration tests, this change should not result in a change in how the integration tests work overall.

It might mean however that changes that before would have not been picked up by the integration tests now will be.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "wdio-selenium-standalone-service": "^0.0.10",
     "wdio-spec-reporter": "^0.1.5",
     "wdio-static-server-service": "^1.0.1",
-    "wdio-webpack-dev-server-service": "^2.0.2",
     "webdriverio": "^4.14.0",
     "webpack": "^4.25.1",
     "webpack-cli": "^3.1.2",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "standard": "standard",
     "test": "run-p standard karma wdio",
     "version": "npm run build && git add -A dist",
-    "wdio": "cross-env NODE_ENV=test SAUCE_CONNECT_VERSION=4.5.4 wdio test/wdio.config.js && git checkout dist"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,7 +27,8 @@
     "standard": "standard",
     "test": "run-p standard karma wdio",
     "version": "npm run build && git add -A dist",
-    "wdio": "cross-env NODE_ENV=test wdio test/wdio.config.js && git checkout dist/"
+    "wdio:test": "cross-env NODE_ENV=test wdio test/wdio.config.js",
+    "wdio": "run-s build wdio:test && git checkout dist/"
```
36degrees, line 31: Is there a reason to use `run-s` here over `npm run build wdio:test && git checkout dist`?

lfdebrux: Ah, I didn't realise that `npm run` could run multiple commands itself. I will update this line.

36degrees: Argh, I think I might be talking rubbish. My brain did not parse that correctly, I was reading it as `wdio:test` being an argument to `npm build` rather than `build` and `wdio:test` being two different tasks 🤦

lfdebrux:
These commits were merged in #450

PR #453, opened 2020-07-06
Add className attribute to status component

lennym:
The status component is currently hidden from view using inline CSS, however in an implementation that applies aggressive CSP controls these styles cannot be applied, and since the status element has minimal other identifying attributes, it is difficult to apply corresponding styles in .css files.

Apply a class attribute to the status component - defaulting to `autocomplete__status` so that it can be reliably targeted by CSS selectors.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -431,6 +431,7 @@ export default class Autocomplete extends Component {
     const wrapperClassName = `${cssNamespace}__wrapper`
 
     const inputClassName = `${cssNamespace}__input`
+    const statusClassName = `${cssNamespace}__status`
     const componentIsFocused = focused !== null
     const inputModifierFocused = componentIsFocused ? ` ${inputClassName}--focused` : ''
     const inputModifierType = this.props.showAllValues ? ` ${inputClassName}--show-all-values` : ` ${inputClassName}--default`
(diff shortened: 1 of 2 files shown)


JakeChampion:
There are a couple of other places where inline styles are being used:
- https://github.com/alphagov/accessible-autocomplete/blob/935f0d43aea1c606e6b38985e3fe7049ddbe98be/src/autocomplete.js#L528
- https://github.com/alphagov/accessible-autocomplete/blob/935f0d43aea1c606e6b38985e3fe7049ddbe98be/src/wrapper.js#L56

@lennym, would you want these to have a class-name defined as well?

lennym:
To be honest, this is a year old and so the problem I was trying to solve has long since left my memory.

Feel free to use your judgement as to what you think is best.

romaricpascal:
@lennym Thanks for adding this. Seems the branch needs a little rebase and for `dist` to be rebuilt. Given your last comment hinting that it had been a while since it was open (and even longer since 😬 ), would you prefer I take care of that (force pushing on your branch, or starting a new PR from a copy of your branch, depending on what GitHub's permissions allow me)?

lennym:
My previous comment still stands... I haven't used this module in literally years, and probably won't in the foreseeable future. 

Do as you see fit.

romaricpascal:
Closing as I've had to create a new branch I could push to. The contents of this PR will be merged by #620.

Issue #460, opened 2020-07-24
Update dependencies


36degrees:
## What

- Bump webpack-dev-server from 3.1.10 to 3.1.11
- Bump mixin-deep from 1.3.1 to 1.3.2
- Bump js-yaml from 3.12.0 to 3.13.1
- Bump handlebars from 4.0.12 to 4.7.6
- Bump acorn from 5.7.3 to 5.7.4

## Why

Dependabot has raised a number of PRs against this repo:

#455 
#456 
#457 
#458 
#459 

Given how long the tests take to run, and the fact that every open PR is automatically rebased when one of them is merged, it's quicker to raise a single PR that bumps all of the outdated dependencies.

PR #462, opened 2020-07-27
Bump dependencies

vanitabarrett:
Resolves https://github.com/alphagov/accessible-autocomplete/issues/460

## What
Bump webpack-dev-server from 3.1.10 to 3.1.11
Bump mixin-deep from 1.3.1 to 1.3.2
Bump js-yaml from 3.12.0 to 3.13.1
Bump handlebars from 4.0.12 to 4.7.6

## Note
Was having some difficulty updating acorn, so going to bump these packages for now and then address that separately.


vanitabarrett:
Good spot @36degrees - I've updated the commit message, as I think we seem to be ok updating to 13.14.0

PR #468, opened 2020-10-09
Update Travis config to paralellize builds

lfdebrux:
I think it would be good to have a Travis build for each test suite, rather than a single Travis build that tests everything. I think this would help in two ways:

1. It should result in more reliability; I've noticed that tests often fail because Travis runs too many jobs at once and SauceLabs only allows 5 connections at a time on the open source plan [[1]]. My hypothesis is that this is because Travis is running the tests twice (once as branch and once as a pull request), and so we're exceeding the SauceLabs quota. Having a job for each test suite means we can skip the SauceLabs tests if it isn't a pull request test.

3. It will result in quicker feedback if tests fail. It might also result in a small overall *increase* in time taken to run all tests, but I think this is less of a problem if you can restart the test that failed only.

[1]: https://saucelabs.com/solutions/open-source

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,12 @@ sudo: required
 dist: trusty
 before_install:
   - export DISPLAY=:99.0
+  - export CXX=g++-4.8
   - sh -e /etc/init.d/xvfb start
   - sudo apt-get update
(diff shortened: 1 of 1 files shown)


lfdebrux:
We're no long using Travis, closing this PR.

Issue #430, opened 2020-04-14
We cannot currently respond to issues or pull requests.


36degrees:
You can still raise issues or pull requests if you want to. As soon as we’re able to, we’ll prioritise dealing with any bugs that have been raised by people in the UK public sector. We cannot prioritise adding new features.

PR #491, opened 2021-05-21
Call the supplied `source` function when the query is empty

JakeChampion:
This fixes the bug report in [this issue comment](https://github.com/alphagov/accessible-autocomplete/issues/210#issuecomment-845860595).

---
## Solution

The solution to this bug is to call the `source` function when the query becomes empty and then let the `source` function decide whether to show no results (closing the menu) or to show some results

---

## Explanation of the bug

This fixes a bug for using async source functions, below is the steps to show the original bug:
1. Create accessible-autocomplete instance with an async source function
2. Type a character and then delete the character really quickly

A live demo of this can be found at <https://output.jsbin.com/werupanune>
You can view the code for the demo at <https://jsbin.com/werupanune/edit?html,js,output>

Accessible-autocomplete would only call the supplied `source` function when the user has enter a value into the text-input and not when the user has deleted all values from the text-input. Below is the part of the code which implements that logic:
https://github.com/alphagov/accessible-autocomplete/blob/935f0d43aea1c606e6b38985e3fe7049ddbe98be/src/autocomplete.js#L226-L241

Because of logic to only call when there is text entered and not when the text has been deleted, there is a race-condition between the async `source` function and accessible-autocompletes own code to close the menu when the query is empty. Accessible-autocomplete will close the menu immediately, but the async `source` function, once it gets it's results, will call the callback to populate the menu with the results from the previous input/query the user had entered.

Below is a video showing the bug:

https://user-images.githubusercontent.com/1569131/119124363-7a679900-ba28-11eb-8028-ecf7ce89d7d3.mov


joelanman:
@JakeChampion Really appreciate your work, however at the moment we don't have capacity to deal with issues and pull requests on this repo unfortunately:
https://github.com/alphagov/accessible-autocomplete/issues/430

Possibly in time we will be able to, and in the meantime I'm sure your work can help people looking for a fix.

edwardhorsford:
I think our team came across this same bug recently.

andygout:
Hi @romaricpascal @colinrotherham @joelanman,

We've noticed there's been more activity on this repo in the last few months and wonder if it might be possible for this PR to be revisited?

([Financial-Times](https://github.com/Financial-Times) maintains a fork of this repo purely to include this bugfix.)

PR #496, opened 2021-06-14
Handle query change when the text length does not change

archferns:
We have encountered a bug wherein a query text change that does not affect its length, does not trigger a search.
This PR aims to fix that by checking that the current and new query texts are not matching, rather than relying on the `length` check.


JakeChampion:
This looks to solve the same issue as https://github.com/alphagov/accessible-autocomplete/pull/491 - is that right?

archferns:
I think the issue might be slightly different. Attaching a recording from our discovery:

The premise here being
1. searching for "college.school", doesn't return any results.
2. searching for "college school", we return a few results.

For a user changing their query from "college.school" to "college school", by selecting the "." character and replacing it with a space key (without using a delete/backspace), we would expect to see a set of results, since the query has changed, however, since the current implementation compares the length of the existing query, and the new query, the autocomplete fails to recognise that the change.


https://user-images.githubusercontent.com/910019/122594083-75fdc280-d05e-11eb-93d6-48ea909c5c11.mp4

romaricpascal:
Hi @archferns, thanks for contributing this fix. Looking to merge it as the code looks good, but it seems that there are some conflicts on the built files. It likely just needs a little rebasing and re-building. Please let me know if you're up for it or if you prefer I do it (either force-pushing on your branch, if GitHub lets me, or opening a new PR that picks your commit) 😊

archferns:
@romaricpascal thank you for that, happy for you to take over!

romaricpascal:
@archferns GitHub wouldn't let me push to your branch, so I've opened #621 for the rebasing and building `dist`. I'll close this one. The code looks good and is tested, so it'll get in soon 🥳

Issue #361, opened 2019-09-18
Failing AXE tests


spacedmonkey:
I used the tool [axe](https://github.com/dequelabs/axe-core) to scan this autocomplete and got the following errors. 

```
    Expected page to pass Axe accessibility tests.
    Violations found:
    Rule: "aria-input-field-name" (ARIA input fields have an accessible name)
    Help: https://dequeuniversity.com/rules/axe/3.3/aria-input-field-name?application=axe-puppeteer
    Affected Nodes:
      #amp-stories-font-family-picker-1__listbox
        Fix ANY of the following:
        - aria-label attribute does not exist or is empty
        - aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
        - Element has no title attribute or the title attribute is empty
    
    Rule: "aria-required-children" (Certain ARIA roles must contain particular children)
    Help: https://dequeuniversity.com/rules/axe/3.3/aria-required-children?application=axe-puppeteer
    Affected Nodes:
      #amp-stories-font-family-picker-1__listbox
        Fix ANY of the following:
        - Required ARIA child role not present: option
```

spacedmonkey:
Another 
```
    Affected Nodes:
      #amp-stories-font-family-picker-5
        Fix ALL of the following:
          - Invalid ARIA attribute value: aria-activedescendant="false".
```

nickcolley:
Hello @spacedmonkey,

I think the error represented by the rule 'aria-input-field-name' will be resolved by the the [the upcoming 2.0.0 release](https://github.com/alphagov/accessible-autocomplete/milestone/3).

When running axecore 3.3.1 on https://alphagov.github.io/accessible-autocomplete/examples/ I cannot reproduce the rule: "aria-required-children", could you help me reproduce that?

The `aria-activedescendant` issue seems to be that the value should always be the ID of an another element, and sometimes it gets set to false, so potentially the logic should be to remove the attribute instead of setting it to false.

robertparkinson:
Another failing aXe issue we have discovered is "Document has active elements with the same id atrribute". The containing div and the input have the same ID.

sussexrick:
@nickcolley We are using v2.0.3 and the `aria-input-field-name` issue is still there. Since this component asks us to render our own label outside this component, I think the fix is to provide a prop where we can pass in the id of that label, and that gets added as the value of `aria-labelledby` on the `<ul>`.

[...]

PR #500, opened 2021-07-22
Fix AXE error: Input fields must have accessible name

sussexrick:
Adds a new `ariaLabelledBy` property which allows consumers to tie the results list to the field label. This improves screen-reader output (tested with NVDA and Chrome/Edge) and allows the code to pass the AXE rule `"aria-input-field-name" (ARIA input fields have an accessible name)`. See #361.


romaricpascal:
Thanks for your contribution. A more generic solution has been implemented in https://github.com/alphagov/accessible-autocomplete/pull/591 for configure that attribute, so I'll be closing this PR.

PR #505, opened 2022-01-05
Fix tests

lfdebrux:
Tests for this repo are currently a bit broken, let's try and fix them.

This PR bundles up a bunch of different changes, but they have to go together because otherwise the tests won't run:

- The SauceLab tests weren't communicating with the service properly (see https://github.com/alphagov/design-system-team-internal/issues/560#issue-1094316650) when run locally, so we bump the `saucelabs` package to fix that...
- but that also requires a bump to the version of Node used (via an indirect dependency on `got`)...
- and after all that we needed to fix Travis, which was just not running at all, with a bunch of apt-get errors.

Closes https://github.com/alphagov/design-system-team-internal/issues/560.


36degrees (approved):
👏🏻 👏🏻 👏🏻

PR #507, opened 2022-01-05
Bump dependencies

lfdebrux:
Bundles all open Dependabot bumps into one PR, to make merging quicker and easier.

Includes changes from the following:

- #465 
- #476 
- #484 
- #486 
- #487 
- #488 
- #492 
- #493 
- #499 
- #501 
- #502 
- #506

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5566,9 +5566,9 @@
       "dev": true
     },
     "dns-packet": {
-      "version": "1.3.1",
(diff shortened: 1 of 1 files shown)


36degrees (approved):
I've had a cursory look through the diffs for each package and they generally look sensible – approving on the basis that the tests still pass and there are no major version changes.

PR #508, opened 2022-01-06
Bump version of Chromium used for functional tests

lfdebrux:
Bumps karma (test runner used for functional tests) and puppeteer, so that we can use a more modern version of Chromium.

This bumps the version of Chromium used for functional tests from 83 to 97.


36degrees (approved):
👍🏻

PR #510, opened 2022-01-06
Use GitHub Actions

lfdebrux:
We want to migrate from using Travis CI to GitHub Actions.

This PR replaces our Travis CI config with two GitHub Actions workflows, `test.yaml` and `saucelabs.yaml`.

Closes https://github.com/alphagov/design-system-team-internal/issues/538.

---

Outstanding:

- [x] Add Sauce Labs secrets
- [x] Tell Travis to stop running on this repo
- [x] Update main branch protection required status checks


.github/workflows/saucelabs.yaml
```diff
@@ -0,0 +1,40 @@
+name: Integration tests (via Sauce Labs)
```
36degrees, line 1: Do we need to do anything to handle PRs from forks, where the sauce labs secrets will not be available?

.github/workflows/saucelabs.yaml
```diff
@@ -0,0 +1,40 @@
...
+        run: npm run build
+
+      - name: Run integration tests
+        env:
+          SAUCE_ENABLED: true
+          SAUCE_BUILD_NUMBER: "${{ github.run_id }}"
+          SAUCE_USERNAME: "${{ secrets.SAUCE_USERNAME }}"
+          SAUCE_ACCESS_KEY: "${{ secrets.SAUCE_ACCESS_KEY }}"
+        run: |
+          npm run wdio:test
```
36degrees, line 40: Any reason this is a block? Given it's a single command, it'd be more consistent to put it on the same line.

```suggestion
        run: npm run wdio:test
```

.github/workflows/saucelabs.yaml
```diff
@@ -0,0 +1,40 @@
+name: Integration tests (via Sauce Labs)
+
+# Our Sauce Labs account only allows 5 tests to run at a time
+concurrency: saucelabs
+
+on:
+  pull_request:
```
36degrees, line 7: Somewhat personal preference, but I find the 'key with nothing after it' slightly strange, and prefer to use [the list style](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-using-a-list-of-events) unless one of the events needs config:

```suggestion
on: [pull_request]
```

.github/workflows/test.yaml
```diff
@@ -0,0 +1,36 @@
+name: Test
+
+on:
+  pull_request:
+  push:
```
36degrees, line 5:
```suggestion
on: [pull_request, push]
```

36degrees (commented):
Looking good, great work! Just a few minor comments.

.github/workflows/saucelabs.yaml
lfdebrux: Ah, good point. I think the way to handle this is to use the `pull_request_target` event [[1](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/), [2](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target)], although apparently that has some associated risks 🤔

lfdebrux: So I think we have three or four options:

1. Run Sauce Labs tests for PRs from forks using `pull_request_target`, and we accept the risk of abuse (similar to what we're doing currently with Travis?
2. Don't run Sauce Labs tests for forks, and don't require them for merge (but developers could still run them locally if they wanted)
3. Run Sauce Labs tests after merge, have a notification to the team if they break, and a policy of fixing any breakages ourself (could be combined with 1)
4. Use a environment for Sauce Labs tests, and require approval for every run

Of course probably the easiest way to get around this would be to remove the Sauce Labs tests altogether 😅

36degrees: Yeah, reading https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ I don't think the `pull_request_target` event is really designed for this.

I think in the past we just didn't run the cross-browser tests on PRs (which also means you can't make it a required status check) – this does mean it's possible to introduce failing tests on main… but as long as that's an accepted risk and we deal with it by either reverting or fixing before a release goes out, that's probably OK?

36degrees: Looks like we commented at the same time – I reckon we're on the same wavelength with 2/3 though and that'd be my preferred approach.

lfdebrux: > I think in the past we just didn't run the cross-browser tests on PRs (which also means you can't make it a required status check)

That wasn't my understanding... 😅 How is that configured?

.github/workflows/saucelabs.yaml
lfdebrux: I found it made it easier to separate it visually from the big block of env declarations above, but I'm not too fussed.

.github/workflows/saucelabs.yaml
36degrees: I think it works because `SAUCE_ENABLED` is a secret in Travis, so is not set on PRs from forks – and the WDIO config is [set conditionally based on the presence of it](https://github.com/alphagov/accessible-autocomplete/blob/master/test/wdio.config.js#L18).

I think this means that on PRs from forks the WDIO tests still ran, but locally instead of using SauceLabs.

.github/workflows/saucelabs.yaml
36degrees: Could you flip `env` and `run` around? It's definitely not a deal breaker – happy to leave as is, just a potential source of confusion for folks that aren't as familiar with YAML.

.github/workflows/saucelabs.yaml
lfdebrux: Ah, I see. So to get the same behaviour, we would just need to move `SAUCE_ENABLED` to be a secret as well, and also make the saucelabs workflow run on updates to main?

lfdebrux: Or make it more explicit, so SAUCE_ENABLED is only true if not running from a fork.

lfdebrux: Okay, I've made it so that SAUCE_ENABLED depends on whether SAUCE_USERNAME is present (which is the easiest way to find if running from a fork that I could find). Tested it both with SAUCE_USERNAME set and unset:

- SAUCE_USERNAME set https://github.com/alphagov/accessible-autocomplete/runs/4739107784?check_suite_focus=true
- SAUCE_USERNAME unset https://github.com/alphagov/accessible-autocomplete/runs/4739082768?check_suite_focus=true

Behaviour was as expected, with tests using Sauce Labs when possible and falling back to Chromium otherwise.

.github/workflows/saucelabs.yaml
```diff
@@ -0,0 +1,50 @@
...
+          node-version: "${{ steps.nvm.outputs.NVMRC }}"
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: "Run integration tests ${{ env.SAUCE_ENABLED == 'true' && '(using Sauce Labs)' || '(using local headless browser)' }}"
```
36degrees, line 45: Nice 👍🏻

.github/workflows/saucelabs.yaml
```diff
@@ -0,0 +1,50 @@
...
+        run: npm run wdio:test
+        env:
+          SAUCE_BUILD_NUMBER: "${{ github.run_id }}"
```
36degrees, line 48: Good idea to move this here – felt a bit clunky when the WDIO config needed to care about the fact it was running on Travis 👍🏻

36degrees (approved):

Issue #420, opened 2020-02-24
Documentation example for templates suggestion:function


DRocksCoding:
Hello,
Is it possible to add a very simple example usage of the templates suggestion with custom html being returned please?

> suggestion is a function that receives one argument, a suggestion to be displayed. It is used when rendering suggestions, and should return a string, which can contain HTML. warning Caution: because this function allows you to output arbitrary HTML, you should make sure it's trusted, and accessible.

Nate-Wessel:
I had a similar thought - displaying the matching text in bold. It doesn't seem like this is possible though because the function isn't passed a query parameter like the `source` function is. Here's an example suggestion function from my code: 
```
suggestion: d => {
	let text = `${d.HScode} - ${d.description}`
	if(d.hasEstimatedGain){ return text }
	return `<span style="color:gray;">${text}</span>`
}
```

vanitabarrett:
We have an existing example for "custom results" in https://alphagov.github.io/accessible-autocomplete/examples/ but we don't have a good code example. We might want to include a realistic code example to show users how it can be used.

Issue #73, opened 2017-04-27
Focus should be unstyled on menu items


edwardhorsford:
Focus should not be styled on menu items. It isn't whilst the item is visibly selected, but if you hover on to another item, focus appears.

Example:
![screen shot 2017-04-27 at 12 44 03](https://cloud.githubusercontent.com/assets/2204224/25482517/205c90b2-2b4a-11e7-8b3d-1d41eb5005a6.png)

36degrees:
This also affects forced colors mode:

<img width="378" alt="Screenshot 2022-01-14 at 10 02 03" src="https://user-images.githubusercontent.com/29889908/149497466-09d6bd9e-7644-47c9-965d-9b5f75bd2a8f.png">

DafyddLlyr:
It would be great to see this issue addressed 👍

PR #512, opened 2022-01-07
Make sure highlighted option is distinguishable in forced colors mode

### LINKED ISSUE #369: Currently highlighted option is not distinguishable in forced color mode
Tested using Firefox settings to change colours. Arrowing down through the results has a couple of problems: a) for the first five results focus isn't visible but the focused result is shown in the input box b) for any following results focus still isn't visible and the input box reverts to only showing the typed in characters, not the full result text, for example "UN" - in the screenshot focus has moved down below the fifth result and the input box just shows UN so the user doesn't know that the result they are on is Egypt. 

screen shot 2017-11-06 at 09 03 02

36degrees:
In forced colours mode (for example, Windows High Contrast Mode) the background and foreground colours of elements are forced to the users preferred colour scheme.

This has the unfortunate side effect of making it impossible to determine which option is currently selected when using the keyboard to move between results in the list, or which item in the list is currently being hovered over.

We can work around this by using the `forced-colors: active` media query (combined with the proprietary `-ms-high-contrast: active` to support IE11) to detect that forced colors mode is active, and then set the background and foreground of the highlighted / hovered option to use the system SelectedItem colour pair (or the Highlight colour pair in browsers that do not yet support SelectedItem).

In Chromium based browsers, a backplate is also being added behind the text for each option. The backplate uses the user's preferred background colour, and ensures that text remains legible if it's drawn over other elements. However, there is no guarantee that there will be sufficent contrast between HighlightText and this background colour – only between HighlightText and Highlight – so we need to disable the backplates within these elements using `forced-color-adjust: none` otherwise they can obscure the text:

Backplate issue

Closes #369.

| Browser | Before | After |
|--|--|--|
| Internet Explorer 11 (Windows HCM, Dark) |  IE11 HCM before (dark) | IE11 HCM After (dark) |
| Internet Explorer 11 (Windows HCM, Light) | IE11 HCM before (light) | IE11 HCM after (light)
| Edge (Windows HCM, Dark) | Edge HCM before (dark) | Edge HCM after (dark) |
| Edge (Windows HCM, Light) | Edge HCM before (light) | Edge HCM after (light) |
| Firefox (custom colours) | Firefox before | Firefox after |


36degrees:
> In Chromium based browsers, a backplate is also being added behind the text for each option. The backplate uses the user's preferred background colour, and ensures that text remains legible if it's drawn over other elements. However, there is no guarantee that there will be sufficent contrast between HighlightText and this background colour – only between HighlightText and Highlight – so we need to disable the backplates within these elements using `forced-color-adjust: none` otherwise they can obscure the text:

Because we've opted out of forced-color-adjust, there is a potential issue that could arise if custom templates are being used that include a child element with a defined text colour. Those child elements will no longer be forced to the user's preferred colour, nor will they be forced to HighlightText, and so we cannot guarantee there will be sufficient contrast between that text and the background.

We _could_ try and prevent this by doing something like:

```css
  .autocomplete__option--focused *,
  .autocomplete__option:hover * {
    background-color: transparent;
    border-color: inherit;
    color: inherit;
  }
```

However, I'm not sure how common this would actually be in practise, so we'd be adding a fair bit of CSS to handle something a hypothetical situation.

Thoughts on how likely this actually is to happen? Any other things we should consider doing to defend against it?

36degrees:
> Because we've opted out of forced-color-adjust, there is a potential issue that could arise if custom templates are being used that include a child element with a defined text colour. Those child elements will no longer be forced to the user's preferred colour, nor will they be forced to HighlightText, and so we cannot guarantee there will be sufficient contrast between that text and the background.
> 
> We _could_ try and prevent this by doing something like:
> 
> ```css
>   .autocomplete__option--focused *,
>   .autocomplete__option:hover * {
>     background-color: transparent;
>     border-color: inherit;
>     color: inherit;
>   }
> ```
> 
> However, I'm not sure how common this would actually be in practise, so we'd be adding a fair bit of CSS to handle something a hypothetical situation.
> 
> Thoughts on how likely this actually is to happen? Any other things we should consider doing to defend against it?

We discussed this as a team and decided that it was not worth guarding against in the component, but that we should document it. I've added a commit to add something to the docs for `templates`.

@EoinShaughnessy would you mind taking a look at it? It's just the change to the README.

README.md
```diff
@@ -268,7 +268,11 @@ This object defines templates (functions) that are used for displaying parts of
 
 `inputValue` is a function that receives one argument, the currently selected suggestion. It returns the string value to be inserted into the input.
 
-`suggestion` is a function that receives one argument, a suggestion to be displayed. It is used when rendering suggestions, and should return a string, which can contain HTML. :warning: **Caution:** because this function allows you to output arbitrary HTML, you should [make sure it's trusted](https://en.wikipedia.org/wiki/Cross-site_scripting), and accessible.
+`suggestion` is a function that receives one argument, a suggestion to be displayed. It is used when rendering suggestions, and should return a string, which can contain HTML.
```
EoinShaughnessy, line 271: Sorry in advance if I've misunderstood - I'm guessing the suggestion is displayed when the user types into a text-field? (Let me know if this is one of those cases where using the passive voice is preferable to writing something contorted.)

Maybe something like this could work:

> `suggestion` is a function that receives one argument, a suggestion for the dropdown to display. The autocomplete uses this function when rendering suggestions. `suggestion` should return a string, which can contain HTML.

README.md
```diff
@@ -268,7 +268,11 @@ This object defines templates (functions) that are used for displaying parts of
...
+
+:warning: **Caution:** because this function allows you to output arbitrary HTML, you should [make sure it's trusted](https://en.wikipedia.org/wiki/Cross-site_scripting), and accessible.
```
EoinShaughnessy, line 273: What does 'arbitrary' mean here? Random?

I get that it's the HTML that needs to be trusted and accessible, but that might not be totally clear at first glance.

README.md
```diff
@@ -268,7 +268,11 @@ This object defines templates (functions) that are used for displaying parts of
...
+
+If your template includes child elements with defined foreground or background colours, you should check that they display correctly in forced colors modes such as Windows high contrast mode.
```
EoinShaughnessy, line 275:
```suggestion
If your template includes child elements with defined foreground or background colours, you should check they display correctly in forced colors modes. For example, Windows high contrast mode.
```

EoinShaughnessy (commented):
Reviewed! :)

README.md
36degrees: I haven't added this line in this PR – just split it up. However, whilst we're here, I agree it could be improved and your suggestion's a really good start.

I think it'd probably make sense to swap the order around – at the minute we talk about what the function takes as input, what it's for, and what it should output… which seems muddled.

What about something like…

> `suggestion` can be used to customise how suggestions are rendered in the dropdown. It must be a function which will be called with one argument, which will be the suggestion. The function must return a string, which can contain HTML.

README.md
36degrees: This line also isn't new, just split up.

It's a bit wordier, but what about…

>  To prevent [cross-site scripting vulnerabilities](https://en.wikipedia.org/wiki/Cross-site_scripting) ensure that the string returned by this function contains only trusted or sanitised input. You should also ensure that any additional HTML markup is accessible.

EoinShaughnessy: Yeah, that sounds good. Maybe just the lightest of tweaks:

> To prevent [cross-site scripting vulnerabilities](https://en.wikipedia.org/wiki/Cross-site_scripting), make sure the string returned by this function contains only trusted or sanitised input. You should also make sure any additional HTML markup is accessible.

(Sorry if this seems finicky - we're encouraged to write 'make sure' (plainer alternative) instead of 'ensure', and remove redundant _that_s.)

README.md
36degrees: The more I pick at this the more I want to rewrite the whole section on `templates` – possibly trying to cover off #420 at the same time.

I feel like that's starting to really stretch the scope of this PR, so how would you feel about me adding the new line in without making changes to the existing content, and creating a new issue or PR to tackle the surrounding docs?

36degrees:
Huh, playing around with HCM disabled I think that's actually the default user agent focus ring. I'm not sure what the expected behaviour is when you do that, but I'm not 100% sure the ring appearing is even intentional… 🤔

edwardhorsford:
@36degrees the browser focus ring appearing is a longstanding bug. Browser focus styles are meant to be disabled on the autocomplete, and the component then take care of styling.

There's a few cases where browser focus styles bleed out.

36degrees:
> @36degrees the browser focus ring appearing is a longstanding bug. Browser focus styles are meant to be disabled on the autocomplete, and the component then take care of styling.
> 
> There's a few cases where browser focus styles bleed out.

It looks like this issue is covered in #73.

README.md
36degrees: @EoinShaughnessy any objections to leaving further improvements to the existing docs out for now?

EoinShaughnessy: @36degrees No objections! Sorry for late reply. Give me a shout whenever we want to tackle the follow-up issue.

vanitabarrett (approved):

vanitabarrett:
@36degrees Think this needs a Changelog entry?

36degrees:
Rebased to bring up to date with the changes in `main`.

EoinShaughnessy (approved):

PR #517, opened 2022-01-10
Upgrade WebdriverIO from v6 to v7

lfdebrux:
We want to use the latest version of WebdriverIO, v7.

This PR updates all WebdriverIO packages at once, as recommended by the migration guide [[1]].

The release notes suggest that most of the changes are related to Typescript [[2]], so there shouldn't be any breaking changes for us to worry about.

[1]: https://webdriver.io/docs/v7-migration
[2]: https://webdriver.io/blog/2021/02/09/webdriverio-v7-released/


36degrees (approved):
Approving on the basis that the tests still pass!

PR #518, opened 2022-01-10
Update browserslist database

lfdebrux:
This doesn't change the output of `npm run build`, but it does stop browserslist complaining at us.


36degrees (approved):
👍🏻

Issue #161, opened 2017-06-06
Is "Select your country" the best micro copy?


stephenkeable-allies:
Minor thought about the label for the fields on the examples, which is currently "Select your country" above a text input field. So seemingly with nothing to "select".
For screen readers the status element would advise the user needs to start typing, however other users don't see or hear that text instruction.
Wonder if it's a good idea to use examples with label of "what's your country" and a hint of "Start typing your country" or similar, to encourage developers to use sensible labels too.
of course you'd like to think people are considering this anyway, but some may copy and paste the element and labels without much thought.

joelanman:
good point - should probably just be "Country" for the purposes of the examples.

In a real service we'd want the copy to fit the context. So on a page that asked "What is your addess?" a fields with "Postcode", "Country", etc might be fine. In another context the question might be something like "What country were you born in?" which is a bit wordy and specific for a page of multiple examples.

stephenkeable-allies:
Main issue was the word "select" when it's an element you "type" in.
Have you guys done any research on whether the user needs to be told that results will appear as they start typing, before they interact with the field?
I've seen cases where users who look at the keyboard while typing, end up typing the entire name, not knowing they could have selected the item they wanted after a few keystrokes.

edwardhorsford:
This isn't the language we were using in user research. We had context appropriate questions.

@stephenkeable-allies We've not tried explicitly telling the users that results will appear, but have seen no issues with users inputting country information. We're asking for a piece of information they know, so as long as they know a name for it, it works well for them.

We'll blog a bit about what we found in the research after Purdah. Broadly users fell in to two categories: those that typed two or three characters and then picked a country from suggestions, or those that typed the entire first word of a country. Rarely did users type a full country name containing two or more words. 

As long as they're able to get to a country , we don't much mind how they use it - there's no reason to push them towards 'searching' - as they know the name of the thing. For contexts where the user might not be as confident, I'd agree it would be good for the text to give a hint that you can search.

stephenkeable-allies:
@edwardhorsford very useful as ever, thanks. Having open government departments researching and sharing knowledge is incredibly useful. Look forward to the blog about the research, sounds like users must look up after the first word, which is interesting.

tvararu:
Hello! 👋 

I wrote the content of the `<label>` and can confirm that no thought whatsoever went into it! The docs need a lot of work. The crude examples were hastily put together to demonstrate the API and provide a playground for manual and automated testing.

[...]

PR #519, opened 2022-01-10
Add generic microcopy to autocomplete examples

EoinShaughnessy:
Fixes [#161](https://github.com/alphagov/accessible-autocomplete/issues/161).

This PR replaces the microcopy 'Select your country' with 'Country', above our [autocomplete examples](https://alphagov.github.io/accessible-autocomplete/examples/).

We're making this change because, in a text input field, there would be nothing to 'select'.

Also, in a real service, this microcopy would be context-specific. So, it make sense to keep our own examples generic.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -59,7 +59,7 @@
 
       <h3><code>{ element, id, source }</code></h3>
       <p>This illustrates usage with only the required arguments.</p>
-      <label for="autocomplete-default">Select your country</label>
+      <label for="autocomplete-default">Country</label>
       <div id="tt-default" class="autocomplete-wrapper"></div>
 
       <h3><code>{ minLength: 2 }</code></h3>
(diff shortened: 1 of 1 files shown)


edwardhorsford:
It's long bugged me that one of the examples is a country picker. There's a separate component for that, so it's an odd one to include.

For a country picker you'd likely enhance from a select - that's where `Select your country` came from. For most services I think it would be unlikely they'd enhance from an input.

36degrees:
@edwardhorsford would you mind raising a separate issue about that? I don't disagree, but that would be a much bigger change which I don't think should block this.

edwardhorsford:
> @edwardhorsford would you mind raising a separate issue about that? I don't disagree, but that would be a much bigger change which I don't think should block this.

Raised in #521

vanitabarrett (approved):

EoinShaughnessy:
@vanitabarrett Thanks for reviewing! Sorry for asking - can I get you to take a final look? Just want to be sure I've displayed lines 143-146 correctly.

examples/index.html
```diff
@@ -140,13 +140,15 @@ <h3>Progressive enhancement</h3>
 
       <h3>Custom results</h3>
       <p>
-        This example demonstrates how to allow users to search for a country by either
-        its english name or its endonym. The suggestions are user-defined objects and
+        This example demonstrates how to allow users to search for a country by either:
```
vanitabarrett, line 143: Thanks for flagging this bit specifically, @EoinShaughnessy , I think I missed it in my first review.

Unfortunately we can't do the list like this because this file is written in HTML, not Markdown.

Instead, it'll need to look like this:

```
<p>This example demonstrates how to allow users to search for a country by either:</p>
<ul>
    <li>its English name</li>
    <li>the name by which a group of people refer to their country</li>
</ul>
<p>The suggestions are user-defined objects and they are displayed using a user-defined template. Confirming a suggestion will populate the input with the English country name.</p>
```

EoinShaughnessy: Ahh I see - thanks! That's v useful to know, have noted for future reference. Will add in.

Issue #466, opened 2020-09-24
Docs for autoselect are incorrect


edwardhorsford:
The docs state that autoselect defaults to `false` - this does not appear to be the case at least for the `accessibleAutocomplete.enhanceSelectElement()` method is used - the feature is on by default.

This is problematic as autoselect is basically incompatible with show all values and causes usability issues.

matthewford:
@edwardhorsford im writing a blog post on this atm with a fix in our branch

matthewford:
https://github.com/alphagov/accessible-autocomplete/compare/master...bitzesty:select-behaviour

36degrees:
It looks like the default value for autoselect is `false`:
https://github.com/alphagov/accessible-autocomplete/blob/d617d5f0c60871ad9e0a3000ce8802edad749e92/src/autocomplete.js#L39

However, `enhanceSelectElement` specifically enables it (unless it's passed in as an option):
https://github.com/alphagov/accessible-autocomplete/blob/d617d5f0c60871ad9e0a3000ce8802edad749e92/src/wrapper.js#L45

It [currently says that it will do this in the README](https://github.com/alphagov/accessible-autocomplete#progressive-enhancement) as well:

> You can use the `accessibleAutocomplete.enhanceSelectElement` function to enhance it into an autocomplete:
> 
> ```js
> accessibleAutocomplete.enhanceSelectElement({
>   selectElement: document.querySelector('#location-picker')
> })
> ```
> 
> This will:
> 
> - Place an autocomplete input field after the specified `<select>`
> - **Default the autocomplete `autoselect` to `true`**
> - [...]

So I believe the docs are currently correct, even if the behaviour isn't desired.

PR #523, opened 2022-01-11
Add correct default behaviour for autoselect

EoinShaughnessy:
Fixes [#466](https://github.com/alphagov/accessible-autocomplete/issues/466).

This PR corrects the README's previous wording, which said `autoselect` defaults to `false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,11 +175,11 @@ function suggest (query, populateResults) {
 
 ### Other options
 
-#### `autoselect` (default: `false`)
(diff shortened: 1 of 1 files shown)


36degrees:
After a bit of digging I think this might be more complicated than we first thought – see https://github.com/alphagov/accessible-autocomplete/issues/466#issuecomment-1010165024

EoinShaughnessy:
This suggested update does not seem to be accurate, so we may as well close this PR.

PR #436, opened 2020-06-17
Fix Travis CI config to de-duplicate addons key

lfdebrux:
This might help fix #435?

Issue #224, opened 2017-08-29
Make clear how this repo relates to govuk-country-and-territory-autocomplete


edwardhorsford:
I don't think it's clear how this repo relates to @alphagov/govuk-country-and-territory-autocomplete . Both should be clear.

I've come across two cases of GDS teams using the accessible autocomplete when they probably wanted to be use the country and territory autocomplete. In both cases it seems the teams were not aware of the country and territory autocomplete. This is a massive shame as much of our work / user research / ui tweaks are in the other repo, not this one.

Can we make the differences and the link between them clearer?

vanitabarrett:
There is a similar card in the govuk-country-and-territory-picker repo to make the link with the accessible autocomplete clear. It would make sense to look at both of these together. https://github.com/alphagov/govuk-country-and-territory-autocomplete/issues/23

EoinShaughnessy:
Just wondering about the difference between the autocompletes - is the accessible autocomplete for everything else not covered by the country-and-territory autocomplete?

Re fixing this issue and [#23](https://github.com/alphagov/govuk-country-and-territory-autocomplete/issues/23) - maybe we could:

- be explicit in each README about what the repo at hand is for
- add links from each README to the README for the other autocomplete

edwardhorsford:
@EoinShaughnessy 

There might be an argument for a longterm plan to merge them.

The history of the difference is that we didn't start building a generic component, we started with the country picker - as that was by far the common need for an autocomplete. So by doing a country picker we would get the autocomplete for free.

We did all of our development, design, and user research using the country picker. Most of the deisgn optimisations happened in it. We also made several optimisations specific for picking countries - more complex synonym support and way that hints work.

A subset of the codebase was then forked out to make the accessible autocomplete. We intended that some of the better features from the country picker would be reintroduced, but this sadly never happened. In particular the autocomplete ships with a really naive matching algorithm and no sorting algorithm.

EoinShaughnessy:
@edwardhorsford That's really helpful, thanks for sharing that context!

PR #525, opened 2022-01-12
Link autocomplete READMEs and add info on support

EoinShaughnessy:
Fixes [#224](https://github.com/alphagov/accessible-autocomplete/issues/224) and [#436](https://github.com/alphagov/design-system-team-internal/issues/436).

This PR updates the autocomplete README with:

- a definition of the component
- advice to users about when not to use it
- a link to the country-picker README
- information about what support we can provide
- links to pinned support and research issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,8 @@
 # Accessible autocomplete
 
----
+The accessible autocomplete is a component that helps users choose answers from a list you provide. You can also use it to make the answers you get from users more consistent.
 
-**We cannot currently respond to issues or pull requests.**
-
-You can still raise issues or pull requests if you want to. As soon as we’re able to, we will prioritise dealing with any bugs that have been raised by people in the UK public sector. We cannot prioritise adding new features.
-
----
+If you're asking users to provide their country or territory, the [govuk-country-and-territory-autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/main/README.md) might be more appropriate.
 
 [![npm version](https://img.shields.io/npm/v/accessible-autocomplete.svg)](http://npm.im/accessible-autocomplete)
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
@@ -21,6 +17,18 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
 - **Compatibility**: Working with [recommended browsers](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in) and [assistive technologies](https://www.gov.uk/service-manual/technology/testing-with-assistive-technologies#which-assistive-technologies-to-test-with).
 
 [Try out the examples!](https://alphagov.github.io/accessible-autocomplete/examples/)
+
+---
+
+## Support
+
+The GOV.UK Design System team maintains the accessible autocomplete as a standalone component. However, we’re only able to put in minimal work to support it.
+
+[Read about our plans to maintain this component](https://github.com/alphagov/accessible-autocomplete/issues/532).
+
+[Read more about the types of support we can provide](https://github.com/alphagov/accessible-autocomplete/issues/430).
+
+---
 
 ## Installation / usage
 
@@ -391,7 +399,7 @@ accessibleAutocomplete.enhanceSelectElement({
 })
 ```
 
-Any null options will also be filtered out of the options used to populate the `source` of the autocomplete element. To preserve options with no value in the autcomplete then pass a `preserveNullOptions` flag of `true` to `enhanceSelectElement`.
+Any null options will also be filtered out of the options used to populate the `source` of the autocomplete element. To preserve options with no value in the autocomplete, then pass a `preserveNullOptions` flag of `true` to `enhanceSelectElement`.
 
 ## Analytics and tracking
 


README.md
```diff
@@ -1,5 +1,11 @@
 # Accessible autocomplete
 
+The accessible autocomplete is a component that predicts the rest of a word or string a user is typing.
+
+Do not use the accessible autocomplete for services where users have to declare their country or territory.
```
EoinShaughnessy, line 5: Not sure 'declare' is the right verb, but I thought 'select' might be inaccurate because some users could type out the word in full. Also didn't want to write 'state' in case anyone reads it as a noun and not a verb, or gets confused by seeing 'state' and 'country' in the same sentence...

EoinShaughnessy:
So as not to have multiple PRs floating around, I'm also going to use this PR for the [updated message about support](https://github.com/alphagov/design-system-team-internal/issues/436).

README.md
edwardhorsford: Provide?

README.md
```diff
@@ -1,5 +1,11 @@
 # Accessible autocomplete
 
+The accessible autocomplete is a component that predicts the rest of a word or string a user is typing.
```
edwardhorsford, line 3: I'm not sure about this. It's not just about predicting what they're typing - it's about helping guide users to an answer.

It helps offer suggestions / let users pick from a set of known options. With synonyms it helps find the right answer even when the user uses a different name - that's not the same as predicting what they're typing.

edwardhorsford: It's also about helping standardise responses where people might use different spellings or phrases for the same thing.

EoinShaughnessy: Thanks @edwardhorsford - your suggestions look good. I'm going to combine them with some other changes I want to make (which it would also be great to get your feedback about).

edwardhorsford: @EoinShaughnessy not remotely meant as suggested text! very little thought went in to it ;)

README.md
```diff
@@ -1,12 +1,10 @@
 # Accessible autocomplete
 
----
-
-**We cannot currently respond to issues or pull requests.**
+The accessible autocomplete is a component that helps users choose countries and territories in answers.
```
EoinShaughnessy, line 3: @edwardhorsford Would this be a better description for the country-and-territory autocomplete? Am concerned using the same description in both READMEs risks confusing users.

edwardhorsford: lol no - this is explicitly not the component to use for picking countries and territories 😆

EoinShaughnessy, line 3:
```suggestion
The accessible autocomplete is a component that helps users choose answers from a list we provide.
```

README.md
```diff
@@ -22,6 +20,34 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
 
 [Try out the examples!](https://alphagov.github.io/accessible-autocomplete/examples/)
 
+---
+
+## Support
+
+The GOV.UK Design System team maintains the accessible autocomplete as a standalone component. However, we’re only able to put in minimal work to maintain and support it.
```
calvin-lau-sig7, line 27: We're waiting for broader feedback from the team, however with this wording you might be able to avoid repeating "maintain":

```suggestion
The GOV.UK Design System team maintains the accessible autocomplete as a standalone component. However, we’re only able to put in minimal work to support it.
```

EoinShaughnessy: Thanks! Will make the same change in the PR for the country-picker README.

README.md
```diff
@@ -1,12 +1,10 @@
 # Accessible autocomplete
 
----
-
-**We cannot currently respond to issues or pull requests.**
+The accessible autocomplete is a component that helps users choose countries and territories in answers.
 
-You can still raise issues or pull requests if you want to. As soon as we’re able to, we will prioritise dealing with any bugs that have been raised by people in the UK public sector. We cannot prioritise adding new features.
+Do not use the accessible autocomplete for services where users have to provide their country or territory.
```
calvin-lau-sig7, line 5: Just checking, what would we say if a service team needed to use a custom country/territory list?

Would we tell them adapt the 'country and territory autocomplete' or is 'accessible autocomplete' somehow better for customising?

README.md
```diff
@@ -1,12 +1,10 @@
 # Accessible autocomplete
 
----
-
-**We cannot currently respond to issues or pull requests.**
+The accessible autocomplete is a component that helps users choose answers from a list we provide. You can also use it to standardise user input.
```
EoinShaughnessy, line 3: Do we need to give an example of how it helps standardise input, or is that something our users will understand?

README.md
edwardhorsford: @calvin-lau-sig7 when you use the `country and territory autocomplete` you can provide your own list of countries.

Alternatively (what my team has done) a team could use the default list, but filter out ones they don't want.

README.md
edwardhorsford: Suggest `answers from a list you provide`

README.md
EoinShaughnessy, line 3:
```suggestion
The accessible autocomplete is a component that helps users choose answers from a list you provide. You can also use it to standardise user input.
```

README.md
calvin-lau-sig7: Thanks @edwardhorsford, that makes things clear, cheers!

README.md
```diff
@@ -1,12 +1,10 @@
 # Accessible autocomplete
 
----
-
-**We cannot currently respond to issues or pull requests.**
+The accessible autocomplete is a component that helps users choose answers from a list you provide. You can also use it to standardise user input.
```
EoinShaughnessy, line 3: @36degrees Do we need to give an example of how it helps standardise input, or is that something our users will understand?

README.md
```diff
@@ -1,12 +1,10 @@
...
 
-You can still raise issues or pull requests if you want to. As soon as we’re able to, we will prioritise dealing with any bugs that have been raised by people in the UK public sector. We cannot prioritise adding new features.
+Do not use the accessible autocomplete for services where users have to provide their country or territory.
```
36degrees, line 5: 'Do not use' feels a little over the top here. There might be valid reasons why the country and territory autocomplete doesn't meet a users needs, like the fact it relies on jQuery.

How about…

> If you're asking users to provide their country or territory, the [govuk-country-and-territory-autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/main/README.md) may be more appropriate.

README.md
36degrees: I'm not sure it's clear what we mean by 'standardise user input'. I guess the way the autocomplete does this is to provide _suggestions_ which hopefully users would pick if they matched, but ultimately allow the user to type whatever they want, which is what makes it different from the use case in the first sentence?

README.md
```diff
@@ -22,6 +20,34 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
...
+
+[Read more about the types of support we can provide.](LINK-TO-PINNED-SUPPORT-ISSUE)
+
+We cannot currently provide support for:
+
+- adding new features
+- smaller functionality issues that do not block usage
+- issues caused by users customising the accessible autocomplete
+- issues with browsers and assistive technologies not mentioned in the [GOV.UK Service Manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices)
+- complex issues with React
```
36degrees, line 47: Not 100% sure about the use of 'provide support' here – possibly it implies we can only help users to fix these things, rather than fixing them ourselves? It reads slightly oddly to me – but if others are happy with it then feel free to ignore.

README.md
calvin-lau-sig7: Ahh okay great, @EoinShaughnessy and I tried to see if there were other reasons but did not come up with anything.

EoinShaughnessy: Looks great, thanks both 👍🏻

README.md
EoinShaughnessy: Should we add something like this?

> For example, when the user of a health service types ‘chest’ - they might see an option that fits their need, but using another term. Similarly, you can use synonyms. For example, to display ‘heart pain’ when the user types ‘chest’.

cc @calvin-lau-sig7

README.md
EoinShaughnessy: @36degrees Would it clear up things up at all if we wrote "We support" and "We cannot currently support" instead?

README.md
calvin-lau-sig7: How about saying 'make answers more consistent'?
I'm a bit mindful going into an example might push down the rest of the intro, and if so we'd probably want to look for a better or clearer example.

```suggestion
The accessible autocomplete is a component that helps users choose answers from a list you provide. You can also use it to make the answers you get from users more consistent.
```

edwardhorsford:
>I'm not sure it's clear what we mean by 'standardise user input'. I guess the way the autocomplete does this is to provide suggestions which hopefully users would pick if they matched, but ultimately allow the user to type whatever they want, which is what makes it different from the use case in the first sentence?

@36degrees It depends how it's configured - In some configurations that's right, but in others users are meant to pick from the options only.

README.md
EoinShaughnessy: Looks good to me. What do you think, @edwardhorsford and @36degrees?

README.md
```diff
@@ -22,6 +20,34 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
 
 [Try out the examples!](https://alphagov.github.io/accessible-autocomplete/examples/)
 
+---
+
+## Support
+
+The GOV.UK Design System team maintains the accessible autocomplete as a standalone component. However, we’re only able to put in minimal work to support it.
+
+[Read about our plans to maintain this component.](LINK-TO-PINNED-RESEARCH-ISSUE)
```
EoinShaughnessy, line 29:
```suggestion
[Read about our plans to maintain this component](LINK-TO-PINNED-RESEARCH-ISSUE).
```

README.md
```diff
@@ -22,6 +20,34 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
...
+
+We provide support for:
+
+- accessibility issues that block usage
+- security issues
+- broken functionality that blocks usage for most users
+- dependency updates to keep tests working so we can make changes safely
+- documentation updates
+
+[Read more about the types of support we can provide.](LINK-TO-PINNED-SUPPORT-ISSUE)
```
EoinShaughnessy, line 39:
```suggestion
[Read more about the types of support we can provide](LINK-TO-PINNED-SUPPORT-ISSUE).
```

README.md
EoinShaughnessy, line 29:
```suggestion
[Read about our plans to maintain this component](LINK-TO-PINNED-RESEARCH-ISSUE).
```

README.md
EoinShaughnessy, line 39:
```suggestion
[Read more about the types of support we can provide](LINK-TO-PINNED-SUPPORT-ISSUE).
```

README.md
EoinShaughnessy: Just commenting to flag that this comment thread isn't outdated and we still need to decide on wording.

README.md
```diff
@@ -1,12 +1,10 @@
...
----
-
-**We cannot currently respond to issues or pull requests.**
+The accessible autocomplete is a component that helps users choose answers from a list you provide. You can also use it to standardise user input.
 
-You can still raise issues or pull requests if you want to. As soon as we’re able to, we will prioritise dealing with any bugs that have been raised by people in the UK public sector. We cannot prioritise adding new features.
+If you're asking users to provide their country or territory, the [govuk-country-and-territory-autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/main/README.md) might be more appropriate.
 
----
+To find information about the autocomplete for choosing country and territory, go to the [govuk-country-and-territory README](https://github.com/alphagov/govuk-country-and-territory-autocomplete/blob/main/README.md).
```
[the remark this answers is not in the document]
EoinShaughnessy: Yup! Will do, thanks.

README.md
36degrees: Not sure it changes much TBH. I still can't articulate what's bugging me about it, so happy to be ignored on this one! 🙂

EoinShaughnessy: Fair enough!

36degrees (approved):

README.md
```diff
@@ -22,6 +18,18 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
 
 [Try out the examples!](https://alphagov.github.io/accessible-autocomplete/examples/)
 
+---
+
+## Support
+
+The GOV.UK Design System team maintains the accessible autocomplete as a standalone component. However, we’re only able to put in minimal work to support it.
+
+[Read about our plans to maintain this component](LINK-TO-PINNED-RESEARCH-ISSUE).
```
EoinShaughnessy, line 27:
```suggestion
[Read about our plans to maintain this component](https://github.com/alphagov/accessible-autocomplete/issues/532).
```

README.md
```diff
@@ -22,6 +18,18 @@ You can still raise issues or pull requests if you want to. As soon as we’re a
...
+
+[Read more about the types of support we can provide](LINK-TO-PINNED-SUPPORT-ISSUE).
```
EoinShaughnessy, line 29:
```suggestion
[Read more about the types of support we can provide](https://github.com/alphagov/accessible-autocomplete/issues/430).
```

EoinShaughnessy:
Holding off on merging till the [sibling PR for country autocomplete](https://github.com/alphagov/govuk-country-and-territory-autocomplete/pull/101/files) gets an approval.

PR #543, opened 2022-02-14
Fix integration tests when testing locally (not with Sauce Labs)

lfdebrux:
The tests were failing when not using Sauce Labs, because `selenium-standalone` was downloading a version of ChromeDriver that expected a newer version of Chromium than Puppeteer was installing.

This PR tries to fix things by replacing the use of Selenium/WebDriver with the Chrome DevTools protocol, which webdriver.io (the test runner) supports [[1]]. This means we get to uninstall things, yay!

We also change the GitHub Actions setup so the integration tests using Chrome always get run, even if Sauce Labs is enabled.

[1]: https://webdriver.io/docs/automationProtocols


vanitabarrett (approved):
I don't fully understand some of these changes (specifically commits 2-3), so I'd appreciate some more reviews before this gets merged. However, the other commits all seem sensible, the tests are passing on this PR and I've checked the branch out locally and run the app and the tests locally successfully too, so I'm fairly happy that this all works.

vanitabarrett:
Thanks Laurence. Some of this may be difficult to explain in a commit message - I think I just don’t have enough experience in this area to fully understand some of these things.

However, in case these are helpful, here are the specific questions that come to mind:

Commit 2: The docs suggest this uses the devtools npm package (https://www.npmjs.com/package/devtools) - is that something that comes bundled with WebDriver, as it doesn’t look like we’re installing anything new? Is there a benefit to using this?

Commit 3: From what I understand, we have a test which attempts to click an element which isn’t clickable (for whatever reason) when running the tests, but works when interacted with by a human. Does the click *always* error? Do we get anything valuable from running this test?

lfdebrux:
> Commit 2: The docs suggest this uses the devtools npm package (https://www.npmjs.com/package/devtools) - is that something that comes bundled with WebDriver, as it doesn’t look like we’re installing anything new? Is there a benefit to using this?

`devtools` appears to be a dependency of `webdriverio`, so we don't need to worry about installing it.

```
accessible-autocomplete@2.0.4
├─┬ @wdio/cli@7.16.13
│ └─┬ webdriverio@7.16.13
│   └── devtools@7.16.13
├─┬ @wdio/local-runner@7.16.13
│ └─┬ @wdio/runner@7.16.13
│   └─┬ webdriverio@7.16.13
│     └── devtools@7.16.13
├─┬ @wdio/sauce-service@7.16.13
│ └─┬ webdriverio@7.16.13
│   └── devtools@7.16.13
├─┬ @wdio/sync@7.16.13
│ └─┬ webdriverio@7.16.13
│   └── devtools@7.16.13
└─┬ webdriverio@7.16.13
  └── devtools@7.16.13
```

lfdebrux:
> Commit 3: From what I understand, we have a test which attempts to click an element which isn’t clickable (for whatever reason) when running the tests, but works when interacted with by a human. Does the click always error? Do we get anything valuable from running this test?

Great question. The click always errors when using Selenium and ChromeDriver, but not when using the DevTools protocol. I _think_ the thing we're trying to test is that if you somehow manage to click on the text rather than the list item containing the text, it should act in the same way. In a sense the error 'Other element would receive the click' is good enough to assert that the correct behaviour would be followed.

I don't know enough about frontend to know whether this is test is actually needed; based on what it's doing, it does seem like its testing  normal browser behaviour that you should be able to rely on, but I guess it's possible to break things in wayst that I'm not aware of...

vanitabarrett:
@lfdebrux That makes sense, and thank you for updating the commit messages! I understand it a lot better now, thank you for taking the time to explain 🙂

36degrees (approved):
Looks good to me 👍🏻

PR #553, opened 2022-04-21
ariaActivedescendant = false not valid aria

### LINKED ISSUE #434: Active descendant false fails axe and lighthouse tests
## Repro steps

1. Load accessible autocomplete on a page
2. Run axe

## Expected results

No accessibility issues reported

## Actual results

Axe reports "ARIA attributes must conform to valid values, Ensures all ARIA attributes have valid values, Invalid ARIA attribute value: aria-activedescendant="false".

## Notes

It seems to be cause by this line: https://github.com/alphagov/accessible-autocomplete/blob/master/src/autocomplete.js#L495. Maybe this should return a null value instead of false, so that the attribute isn't even rendered? Or an empty string so it's treated as empty?

This [comment in the ARIA spec:](https://github.com/w3c/aria/issues/501#issuecomment-335679591) suggests having an empty string is OK and is treated as an absent property.

The [ARIA spec](https://www.w3.org/TR/wai-aria-1.1/#aria-activedescendant) says activedescendant can only take an 'ID reference' so a bool value is invalid.

This [ARIA practices example](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html) uses an empty value, rather than removing the attribute.

Here's a [discussion on axe core about activedescendant](https://github.com/dequelabs/axe-core/issues/179).

I don't know what the actual implications of having this invalid attribute are for assistive tech but it does mean that automated accessibility tests fail. In our case we're running them in out CI pipelines to catch any basic accessibility regressions.

Jordan-Hall:
Fixes #434

I've used this as the base code for DLUHC. Needed to make small changes to get it to work with plotly dash so migrated from classes to hooks. This is one of the changes we've made to ensure lighthouse was 100.

PS: Happy to push the react hooks changes too if you are interested.

### DIFF
--- a/src/autocomplete.js
+++ b/src/autocomplete.js
@@ -456,7 +456,7 @@ export default class Autocomplete extends Component {
     const ariaProps = {
       'aria-describedby': ariaHint ? assistiveHintID : null,
       'aria-expanded': menuOpen ? 'true' : 'false',
-      'aria-activedescendant': optionFocused ? `${id}__option--${focused}` : false,
+      'aria-activedescendant': optionFocused ? `${id}__option--${focused}` : null,
       'aria-owns': `${id}__listbox`,
       'aria-autocomplete': (this.hasAutoselect()) ? 'both' : 'list'
     }


romaricpascal:
Hi @Jordan-Hall, thanks for contributing this fix. The changes look fine, but there are a couple of logistics bits that need to happen before we can merge them:
* the branch needs to be rebased on `main`
* `dist` needs to be re-built to grab the changes
* The GitHub workflows need to run (hopefully pushing the first two changes should trigger that)

Would you be up for doing this or would you prefer I did it (appreciate it took us a bit for us to get to review this, so very fine to do it, thanks for your patience, btw).

On the topics of React hooks, thanks for offering, but we don't have any plans to move towards this. But if you have an open repository at DLUHC, this could be a handy place we could look into for making future decisions 😊

Jordan-Hall:
Thank you @romaricpascal I'll update the branch later. I left DLUHC now I was only contracting for them but still keep an eye on the repo I started for them. I completely rewrote alot of this to use hooks without losing any functionality.


https://github.com/communitiesuk/gov-uk-dash-react-components/blob/main/src/lib/fragments/AutoComplete.react.js I hope this useful for you

romaricpascal:
Brill! Thanks for the links and offering to update the branch 😊

colinrotherham:
I've rebased all open PRs with `main` (where we're able to) including this one

We'd caused a few conflicts by merging https://github.com/alphagov/accessible-autocomplete/pull/600 first so that's all sorted now

nicholasgriffintn:
Not related to this pr, but also sort of related, but thanks for getting this going again!

I'm gonna see if there's anything I can add from my fork when this is all done.

colinrotherham (approved):

colinrotherham:
Needs a 2nd approval on this one @romaricpascal if you're happy that https://github.com/alphagov/accessible-autocomplete/pull/553#issuecomment-1888931397 is complete?

romaricpascal (approved):

PR #579, opened 2022-12-20
Update references to cred store

36degrees:
Part of https://github.com/alphagov/design-system-team-internal/issues/618


domoscargin (approved):
👍

Issue #285, opened 2018-07-25
Font is not set on results


joelanman:
Seems like we need a font declaration in the CSS

![image](https://user-images.githubusercontent.com/1132904/43195991-82523e2e-8ffe-11e8-93a3-7a835ac56819.png)

joelanman:
So I guess this is because currently this component is not like the components in the Design System. I think there's an issue there in that it's hard for people to understand this component is different to our others. Possibly this is an issue to look at when we add this to the Design System.

At the least, we could add this information to 'Styling the Autocomplete' in the readme.

colinrotherham:
Hi @joelanman, we use this component alongside the new design system.

The biggest changes were:

**1. Padding and line heights on input/hint/option has changed**
This needed to be aligned with the autocomplete results (40px height on input/hint)

**2. No defaults fonts on li or input anymore**
i.e. Design system doesn't set `font-family: 'nta', Arial, sans-serif;` by default

Plus we switched to using `govuk-media-query()`.

Perhaps the accessible autocomplete should be updated to compile Sass to CSS using the mixins/variables from `govuk-frontend`?

Gist here of our tweaks, would be good to get some feedback:
https://gist.github.com/colinrotherham/6704c869ed6bc687785158006eba2a04

![sample diff](https://user-images.githubusercontent.com/415517/44452606-e0603100-a5ee-11e8-92c2-a0bbcee22b7b.png)

edwardhorsford:
One of our designers had this whilst trying to add the autocomplete to a prototype.

samuelhwilliams:
We just had this as well. The only change we needed (/decided) to make was to add `@include govuk-typography-common();` into the `.autocomplete__wrapper` class.

edwardhorsford:
We've found we needed to set 
```scss
.autocomplete__wrapper * {
  @include govuk-typography-common();
}
```

Not including the `*` does not apply the font to the intput or hint.

I note that @joelanman's example in the first message has the wrong font on the input too - so the entire autocomplete is missing the font, not just the results. The input uses `system-ui` so looks close _enough_ that it's not as easily seen.

cc @samuelhwilliams since you may have the same issue.

PR #598, opened 2023-08-30
Make Accessible Autocomplete a GOV.UK Prototype Kit plugin

joelanman:
- adds a config file to make the JS and CSS available to the kit as a plugin
- adds info to the Readme so people know how to use it as a plugin


joelanman:
are there any changes I need to make to this?

joelanman:
might need guidance or preferably code to style it like gov.uk

joelanman:
Hi is this something I can help with or shall I close it?

colinrotherham:
@joelanman I've rebased this PR for you and we've added it to the "After next" milestone

You'll see one extra commit https://github.com/alphagov/accessible-autocomplete/pull/598/commits/0be7e76a26cad2f85eaed2e3a4640ccccba12bf9 to avoid https://github.com/alphagov/govuk-prototype-kit/issues/1950

Looks good to me

colinrotherham (approved):

joelanman:
@colinrotherham cheers! The only other thing I can think of is that the autocomplete isn't styled to GOV.UK by default (the font is the most obvious difference). I wonder if it's easy to make that happen when it's installed as a plugin, or if we just give people instructions on how to do it.

colinrotherham:
@joelanman Ah good thinking, forgot I [replied to that one](https://github.com/alphagov/accessible-autocomplete/issues/285#issuecomment-414957207) back in the day

We can review whether `GDS Transport` is allowed within `.govuk-form-group` only perhaps

But I've pushed up some of those tweaks to fix #285 too

joelanman:
@colinrotherham nice! Just to check is the intention to

 - apply some non branded but 'nice' default styling
 - add gov.uk styling if the autocomplete is inside a `govuk-form-group` class?

If so, that seems nice, but I think there's a danger people don't use the `govuk-form-group` class.

My idea is to add another css/sass file purely for gov.uk branding. In the kit this could be included by default (via the plugin config) - others could opt in manually (with instructions in the readme). What do you think?

colinrotherham:
@joelanman Yeah exactly that

I've made sure the defaults use Arial and align with GOV.UK Frontend

Prototype Kit plugin stylesheet sounds like a great idea if there's no nice way to opt-in GOV.UK services too

romaricpascal:
Anticipating some breaking changes from Colin making the autocomplete catch-up with GOV.UK Frontend's style, we're going to split them in their own PR we can add to the current release, which already brings some breaking changes.

I think I'd be keen to have the styling currently set through `.govuk-form-group` in its own spreadsheet specific for the Prototype Kit, and listed in the plugin. That being said, I'm also wondering if that's not something we could solve by making sure you can control the fonts inside the component by setting a class on its root element (or a wrapper) and let the values be inherited when necessary.

colinrotherham:
I've moved the CSS tweaks to https://github.com/alphagov/accessible-autocomplete/pull/644

We're still unsure on how best to serve a separate stylesheet (or styles) for GOV.UK services

README.md
```diff
@@ -75,6 +75,47 @@ You can copy the [dist/accessible-autocomplete.min.js](dist/accessible-autocompl
...
+  </label>
+  <select class="govuk-select" id="country" name="country">
+    <option value="England">England</option>
+    <option value="Northern Ireland">Northern Ireland</option>
+    <option value="Scotland">Scotland</option>
+    <option value="Wales">Wales</option>
+  </select>
+</div>
+```
+
```
joelanman, line 101: I think relying on the class is a bit fragile, but if thats the approach I would suggest calling it out, something like:

Note that the class `govuk-form-group` is required to make the autocomplete look like GOV.UK.

Because if they use examples from elsewhere, it won't have it.

joelanman: Maybe it could get a mention here too?

https://github.com/alphagov/accessible-autocomplete?tab=readme-ov-file#styling-the-autocomplete

joelanman:
is there anything I can help with on this?

src/autocomplete.css
```diff
@@ -152,6 +152,15 @@
...
 }
 
+/* Try font "GDS Transport" for GOV.UK form groups */
+.govuk-form-group .autocomplete__hint,
+.govuk-form-group .autocomplete__input,
+.govuk-form-group .autocomplete__option {
+  font-family: "GDS Transport", arial, sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
```
tvararu, line 162: To add to what @romaricpascal [was saying](https://github.com/alphagov/accessible-autocomplete/pull/598#issuecomment-2066549418):

This could instead be part of a separate `autocomplete-govuk.css` (or better named) stylesheet that is referenced in `govuk-prototype-kit.config.json`.

That way autocomplete users can choose from:

- GOVUK styling `autocomplete-govuk.css`
- Default styling `autocomplete.css`
- No styling (DIY)

Also wanted to mention: the historic reason for why the autocomplete doesn't use GOVUK Frontend classes is because it precedes v0.1 of GOVUK Frontend. It's also a nice feature which has allowed it to be adopted in places outside Gov such as FT or Wordpress. So yes, keeping the "Default styling" agnostic is still worth doing. 👍

joelanman: I'm wondering if this might work:
```
.govuk-template .autocomplete__hint,
...
```
 - simpler to have one css file
 - one request is better for performance
 - `.govuk-template` is always present in the Prototype Kit, whereas there may not be a parent `.govuk-form-group`

PR #600, opened 2023-09-26
Fix null aria attributes

mark-roberts-ho:
This is a fix for #434

The implementation prevents false aria values from being included.

I have also changed the **build** command, removing the single quotes which do not work in my environment.

Please let me know who is responsible for version number increments in package.json


romaricpascal:
Hi @mark-roberts-ho , thanks for the fix! Regarding the version number increment, this is something we'll change when we release the next version of the package, so no need to worry about it. 

Looks like GitHub is struggling to trigger its workflows, though. Would you be able to force push your branch with a new SHA for that last commit to see if GitHub, please? I think `git commit --amend -C HEAD` should give your commit a new SHA which you can force push with `git push --force-with-lease` 😊

romaricpascal (approved):
@mark-roberts-ho Our merges of a couple of PRs lately has brought quite a few conflicts, so I'll try to rebase your branch so it can be merged, as the code looks neat! Thanks for the contribution.

PR #612, opened 2023-11-20
Update packages, configs + Node.js 20

colinrotherham:
Similar to https://github.com/alphagov/accessible-autocomplete/pull/316 this PR brings all our dependencies up to date

Excluding two major bumps which I've split out:

1. Upgrade to [webpack v5](https://webpack.js.org/blog/2020-10-10-webpack-5-release/) in #614
2. Upgrade to [WebdriverIO v8](https://webdriver.io/blog/2022/12/01/webdriverio-v8-released/) in #615

**Note:** Updates to Preact have also been excluded since we'll need to update tests separately. For example, we currently [require a mismatched Preact version](https://github.com/alphagov/accessible-autocomplete/blob/92a1c7200eba68e8f8e8f7d0926bb1fabaaf7397/examples/preact/index.html#L63-L64) via https://unpkg.com

## Configs, tests and polyfills

We know the [deprecated `@wdio/sync` package won't install](https://github.com/alphagov/accessible-autocomplete/issues/608) so tests now use async/await instead

But we can remove more (see below) with config updates, especially now [IE 8–10 support is clarified](https://github.com/alphagov/design-system-team-internal/issues/561):

1. Browserslist config updated with IE11+ to match 
2. Babel config presets updated for [`core-js@3`](https://www.npmjs.com/package/core-js) polyfills
4. Babel config transforms for ES3 removed (prevented errors in IE8)
5. Node.js version config **.nvmrc** updated to v20.9.0 `lts/iron`
6. Husky setup moved to new `.husky/pre-commit` config format

## Packages removed

The following packages were unnecessary or unused:

```console
@wdio/sync
copy-webpack-plugin
dotenv
karma-chai
karma-chai-sinon
replace-bundle-webpack-plugin
sinon
sinon-chai
```

With more included in [`@babel/preset-env`](https://www.npmjs.com/package/@babel/preset-env) automatically now or to transform to ES3 for Internet Explorer 8:

```console
@babel/plugin-proposal-class-properties
@babel/plugin-proposal-decorators
@babel/plugin-transform-member-expression-literals
@babel/plugin-transform-modules-commonjs
@babel/plugin-transform-property-literals
```

## Package audit

See results from `npm audit` below

### Before

```console
found 181 vulnerabilities (3 low, 45 moderate, 104 high, 29 critical) in 2141 scanned packages
  run `npm audit fix` to fix 132 of them.
  43 vulnerabilities require semver-major dependency updates.
  6 vulnerabilities require manual review. See the full report for details.
```

[...]

PR #613, opened 2023-11-20
Update all dependencies using Dependabot

colinrotherham:
This PR enables Dependabot updates for npm and GitHub Actions including:

1. Checks run at `monthly` intervals
2. Checks for unstaged changes after build
3. Grouped PRs for "build", "test" and "tooling" dependencies

Previously only security-related updates were included

~Probably wait for https://github.com/alphagov/accessible-autocomplete/pull/612 to merge first~

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+version: 2
+
+updates:
+  # Update npm packages
+  - package-ecosystem: npm
+    directory: /
+    open-pull-requests-limit: 10
+
+    reviewers:
+      - alphagov/design-system-developers
+
+    # Schedule run every Monday, local time
+    schedule:
+      interval: weekly
+      time: '10:30'
+      timezone: 'Europe/London'
+
+    versioning-strategy: increase
+
+    allow:
+      # Include direct package.json updates
+      - dependency-type: direct
+
+      # Include indirect browser data updates
+      # https://caniuse.com
+      - dependency-name: caniuse-lite
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,40 @@ updates:
     directory: /
     open-pull-requests-limit: 10
 
+    # Group packages into shared PR
+    groups:
+      build:
+        patterns:
+          - '@babel/*'
+          - 'babel-*'
+          - 'core-js'
+          - 'csso-cli'
+          - 'preact'
+          - 'source-map-loader'
+          - 'terser-*'
+          - 'webpack'
+          - 'webpack-*'
+
+      test:
+        patterns:
+          - '@wdio/*'
+          - 'chai'
+          - 'devtools'
+          - 'karma'
+          - 'karma-*'
+          - 'mocha'
+          - 'puppeteer'
+          - 'standard'
+          - 'webdriverio'
+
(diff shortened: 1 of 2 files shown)


36degrees:
I'm not sure if it's realistic for us to stay on top of weekly Dependabot PRs – have added an agenda item to our next 'dev catch up' to discuss what we want to do.

.github/dependabot.yml
```diff
@@ -0,0 +1,72 @@
...
+          - 'chalk'
+          - 'cross-env'
+          - 'dotenv'
+          - 'husky'
+          - 'npm-run-all'
+
+    reviewers:
+      - alphagov/design-system-developers
+
+    # Schedule run every Monday, local time
```
36degrees, line 46: Out of date comment – should now run on the [1st of every month](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#:~:text=Runs%20once%20each%20month.%20This%20is%20on%20the%20first%20day%20of%20the%20month.) (I guess that might be a weekend)

.github/dependabot.yml
```diff
@@ -0,0 +1,72 @@
...
+      # https://caniuse.com
+      - dependency-name: caniuse-lite
+
+  # Update GitHub Actions
+  - package-ecosystem: github-actions
+    directory: /
+    reviewers:
+      - alphagov/design-system-developers
+
+    # Schedule run every Monday, local time
```
36degrees, line 68: Also needs updating.

package.json
```diff
@@ -20,6 +20,7 @@
     "build:css": "csso src/autocomplete.css -o dist/accessible-autocomplete.min.css",
     "build:js": "cross-env NODE_ENV=production webpack --progress",
     "build": "run-s \"build:js\" \"build:css\"",
+    "postbuild": "node scripts/check-staged.mjs",
```
36degrees, line 23: Do we really want this to run whenever we run `npm run build` locally? I'm not sure I follow the thinking here.

colinrotherham: Yeah think so, for local development you'll run `npm run dev` instead

From our contributor PRs you'll see everyone misses it currently and follows up with a commit

Output will also stay silent unless `dist/` has changed

36degrees: > From our contributor PRs you'll see everyone misses it currently and follows up with a commit

Won't that still happen, if contributors are using `npm run dev` instead of `npm run build`?

Where are we expecting `npm run build` to be run where we want this to happen? Are there any places in CI where we're not adding `--ignore-scripts`?

I won't block this PR on it, but unless there's a clear reason for this change I'd be tempted to remove this change and have the 'Check for unstaged changes' job call `node scripts/check-staged.mjs` directly.

36degrees (approved):

package.json
colinrotherham: >Won't that still happen, if contributors are using `npm run dev` instead of `npm run build`?

Nah we use [`webpack-dev-server`](https://webpack.js.org/configuration/dev-server/) so output isn't written to disk for `npm run dev`

We use `npm run build` for the "**To build the project for distribution**" docs + commit hook

I'm happy either way so I'll make your changes, not a problem

PR #614, opened 2023-11-21
[WIP] Upgrade to webpack v5

colinrotherham:
Updates to webpack v5 as a single config with multiple entries, rather than an array of multiple configs

Work in progress due to:

1. Externals `react`, `preact` etc
2. Environment `COMPONENT_LIBRARY` switching


webpack.config.mjs
```diff
@@ -105,78 +119,104 @@ const bundleStandalone = {
...
-      commonjs: 'preact',
-      commonjs2: 'preact',
-      root: 'preact'
-    }
+    preact: 'preact'
   },
+
+  output: {
+    ...config.output,
+    globalObject: 'window'
```
romaricpascal, line 167: **issue** I think this one needs to go in the shared config's `output`, so that the wrapped version of the library also gets added to `window`, as it is with Webpack 4. With Webpack 5, looks like it gets added to `self` instead.

romaricpascal: Interestingly, looks like on GOV.UK Frontend, we add to `this` rather than either of `window` or `self`. `this` sounds more portable, but:
- isn't like for like with the current output of Webpack
- I'm not quite sure if there'd be unforeseen side effects

romaricpascal (commented):
Project built OK and accessible-autocomplete.min.js loaded OK in a browser.

Tried to load the library (using Node, for quickness) both with `require` and `import` which noticed that `dist/accessible-autocomplete.min.js` was adding to `self` rather than `window` [as it does now](https://github.com/alphagov/accessible-autocomplete/blob/main/dist/accessible-autocomplete.min.js), so we should probably go like-for-like (or even move to `this` as GOV.UK Frontend does?).

webpack.config.mjs
colinrotherham: @romaricpascal Shall we pair on this?

You'll notice the difference when importing into the Review app

romaricpascal (approved):

webpack.config.mjs
colinrotherham: Yeah, it's when running without a bundler (e.g. dependencies via unpkg `<script>`)

It means we can use `window.React.createElement()` in user code but still ensure React inside our bundled library code still resolves and extends the `window.React` version using webpack `globalObject`

I'll add a comment to explain:

```js
// Support extending `window.React` when not bundled
// e.g. with all dependencies included via unpkg.com
globalObject: 'window'
```

colinrotherham:
> Let's just check that things still work serverside for React, as it was something [specifically handled by this PR](https://github.com/alphagov/accessible-autocomplete/pull/415).

@romaricpascal Yep, all working as discussed on Slack

There's an extra commit https://github.com/alphagov/accessible-autocomplete/pull/614/commits/1746717cac609bdb90eaf76ac88ed63574289531 to fix server-side imports without breaking `window.React`

I've clarified why we use `globalObject` in https://github.com/alphagov/accessible-autocomplete/pull/614#discussion_r1453522399 and have used this snippet to test a "string render" of the component via the React bundle both client-side and server-side:

```js
import Autocomplete from 'accessible-autocomplete/react.js'
import React from 'react'
import ReactDOMServer from 'react-dom/server.js'

console.log(
  ReactDOMServer.renderToString(
    React.createElement(Autocomplete.default, {}, null)
  )
)
```

romaricpascal (approved):
Looks all good to go. Had a quick re-test [through GOV.UK Frontend review app](https://github.com/alphagov/govuk-frontend/compare/accessible-autocomplete-test-boilerplate?expand=1) and it all worked fine.

PR #615, opened 2023-11-21
Upgrade to WebdriverIO v8

colinrotherham:
Upgrades to [WebdriverIO v8](https://webdriver.io/blog/2022/12/01/webdriverio-v8-released/) from v7

There were some regressions in IE11 and Firefox 55, including:

* https://github.com/webdriverio/webdriverio/issues/11705
* https://github.com/webdriverio/webdriverio/issues/11711
* https://github.com/webdriverio/webdriverio/issues/11716

But these are now resolved

## Package audit

See results from `npm audit` below

```console
found 0 vulnerabilities
```


test/integration/index.js
```diff
@@ -1,9 +1,10 @@
-/* global $, afterEach, beforeEach, browser, describe, it */
+/* global afterEach, beforeEach, describe, it */
 
 const { mkdir } = require('fs/promises')
 const { dirname, join } = require('path')
 const { cwd } = require('process')
-const { expect } = require('chai')
+const { $, browser, expect } = require('@wdio/globals')
```
romaricpascal, line 6: **suggestion** Looks like WebDriver provides its own `expect` function, however it has a different syntax for its matchers than `chai`. That means we now need to juggle two syntaxes depending on whether we're writing integration or functional tests. If we swap, I'd be keep to make the jump fully and take advantage of WebDriver's specialised matchers like, [`toBeDisplayed`](https://webdriver.io/docs/api/expect-webdriverio/#tobedisplayed), [`toBeFocused`](https://webdriver.io/docs/api/expect-webdriverio/#tobefocused),...

romaricpascal: Liking that we don't have `$` and `browser` as globals anymore and we know explicitely where they come from, though 😍

test/wdio.config.js
```diff
@@ -6,73 +6,117 @@ require('@babel/register')({
...
+
+    /**
+     * Web server options
+     *
+     * @type {[string, SauceServiceConfig]}
+     */
+    ['sauce', {
+      sauceConnect: SAUCE_ENABLED === 'true'
+    }]
+  ],
```
romaricpascal, line 111: **suggestion** This part tripped me up a bit, as it always sets up a [Sauce service](https://webdriver.io/docs/sauce-service), even when testing without the `SAUCE_...` environment variables. This makes me a bit uneasy, as I could only _think_ that it was not affecting any local tests through:
- `sauceConnect` would be false due to the lack of `SAUCE_ENABLED`
- the lack of `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` environment variables to configure the connection to SauceLabs
- the lack of `sauce:options` in the `localCapabilities` (less sure of that one)

I'd find it safer to not set up the service altogether (unfortunately, requires a `.filter(Boolean) to clean up the `null` value):
```suggestion
    /**
     * WebdriverIO Sauce service options
     *
     * @type {[string, SauceServiceConfig]}
     */
    SAUCE_ENABLED === 'true'
      ? ['sauce', { sauceConnect: true }]
      : null
  ].filter(Boolean)
```

Alternatively, I'd be fine with a comment in the code to explain why this doesn't impact local runs.

romaricpascal (commented):
Update's looking good 🙌🏻 Thinking we can buy into WebDriver a little more if we use it for expecations, and that we may need a little update (either code or comment) around the setup of Sauce Service 😊

test/integration/index.js
colinrotherham: Cheer @romaricpascal

Is it a blocker?

test/wdio.config.js
colinrotherham: Brill, I'll add a comment to say we're using `sauceConnect` as designed

E.g. I changed it after seeing an example use `sauceConnect` to optionally enable or disable Sauce Labs

colinrotherham:
I've temporarily rolled back to `webdriverio@8.27.0` since an IE11 fix of ours was accidentally reverted:

* https://github.com/webdriverio/webdriverio/pull/11990#issuecomment-1893573025

colinrotherham:
@romaricpascal Ready for review again

romaricpascal (approved):
It's green again! Thanks for looking at what needed to be pinned ⛵

colinrotherham:
Let's merge this and let [Dependabot pick up updates](https://github.com/alphagov/accessible-autocomplete/pull/613) once https://github.com/webdriverio/webdriverio/commit/f4f924b1efa3420cad54f01e7d85b49b8e9f2cea is released

PR #618, opened 2023-12-06
Add Composer Support

bbertucc:
[Equalify](http://equalify.app) would love to include accessible-autocomplete via Composer. This file should set up the package. You'll just need to merge it and submit the repo to [the packagist](https://packagist.org/packages/submit).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "accessible-autocomplete",
+    "name": "alphagov/accessible-autocomplete",
     "description": "An autocomplete component, built to be accessible.",
     "version": "2.0.4",
     "keywords": [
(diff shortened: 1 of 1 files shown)


romaricpascal:
Thanks for your contribution. Unfortunately, we're not looking to publish our package to Composer's package repository as we don't have plans to support that package manager. 

That said, you might be able to use our package with Composer already through [their `package` repository type](https://getcomposer.org/doc/05-repositories.md#package-2). I haven't tested it, but it seems such configuration could be pointed to the ZIP file of [our releases](https://github.com/alphagov/accessible-autocomplete/releases/) on GitHub and would let Composer download the package.

PR #632, opened 2024-01-18
Add example of how to get suggestions via AJAX

romaricpascal:
Prompted by the review of #611, adds an example illustrating how to implement fetching suggestions from a remote server. It deals with the different bits of logistics around sending GET requests repeatedly:
- providing user feedback that results are being fetched, including after results are shown
- providing user feedback if the request errors
- debouncing the requests while users type to avoid overloading servers and waste bandwidth with requests whose results will get replaced after the next keystroke
- cancelling in-flight requests to ensure that only the results from the last request sent will show in the result, avoiding a response from an earlier request, but delayed by network and server timings, to replace that of a later request.

https://github.com/alphagov/accessible-autocomplete/assets/396367/8ad030ac-5b5b-4fc6-af9f-64b687d436d1


tomayac:
This PR is gold and should be merged. Not showing a dynamic `fetch()` is my pet peeve with any autocomplete component.

36degrees (approved):

PR #641, opened 2024-01-19
Fix various Karma + webpack warnings and minor issues

colinrotherham:
This PR includes a few follow-up fixes to Karma + webpack

### Fix Karma webpack output to prevent overwriting `dist/`

Previously `npm run karma` used webpack to overwrite `dist/` but using [`karma-webpack` defaults](https://github.com/codymikol/karma-webpack#default-webpack-configuration) instead. This has now been fixed so that Karma output is written to `os.tmpdir()` as the author intended

### Fix Karma webpack `karma-sourcemap-loader` warnings

Karma webpack `devtool` option now [outputs inline source maps](https://www.npmjs.com/package/karma-webpack#source-maps) to fix warnings about missing source maps

### Ensure webpack uses new `serve` CLI command

We can now use the webpack CLI rather than the `webpack-dev-server` package binary:

```console
npx webpack
npx webpack serve
```

### Ensure webpack always exits on build errors

The [webpack `NoEmitOnErrorsPlugin` plugin](https://webpack.js.org/plugins/NoEmitOnErrorsPlugin/) is now enabled by default but we turn it off in development mode to help diagnose problems in the browser rather than exit

# Warnings since webpack v5 update

Additionally, new warnings have been fixed following the webpack v5 update:

* https://github.com/alphagov/accessible-autocomplete/pull/614

### Fix Karma webpack performance warnings

The Karma webpack `stats: 'errors-only'` option has been updated for webpack v5 to ignore performance warning recommendations since these are only useful for production builds

### Fix Karma webpack `DefinePlugin` warning for `NODE_ENV`

The Karma webpack build runs with `NODE_ENV=test` but this logs a warning in webpack v5 because “test” does not match webpack `config.mode` allowed values `"none" | "development" | "production"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "build:css": "csso src/autocomplete.css -o dist/accessible-autocomplete.min.css",
     "build:js": "cross-env NODE_ENV=production webpack --progress",
     "build": "run-s \"build:js\" \"build:css\"",
-    "dev": "cross-env NODE_ENV=development webpack-dev-server",
+    "dev": "cross-env NODE_ENV=development webpack serve",
     "karma:dev": "cross-env NODE_ENV=test karma start test/karma.config.js",
     "karma": "npm run karma:dev -- --single-run",
     "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\"",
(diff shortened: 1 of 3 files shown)


romaricpascal (approved):
That's a good couple of tidy ups 🙌🏻 Tried dev, test and build commands locally and it all looked fine. ⛵

PR #644, opened 2024-01-22
Align CSS styles with GOV.UK Frontend

colinrotherham:
This PR makes a few tweaks to use the [GOV.UK Frontend v3+ colour palette](https://designnotes.blog.gov.uk/2019/07/29/weve-updated-the-gov-uk-colours-and-font/) and related size changes

We'll investigate fixes for https://github.com/alphagov/accessible-autocomplete/issues/285 in another PR


romaricpascal (approved):
Nice to have those refreshed. Let's see how the discussion on Slack pans out to see if we want a light opinion on the font-family as well to avoid the unstyled results. Approving already and if things are positive on Slack, you can just add it in before merging. ⛵

PR #649, opened 2024-01-26
Update options for styling component inputs

romaricpascal:
When using the `autoselect` option, the component renders a hint when the value typed by the user matches the start of a suggestion. The styles of this hint need to match some of the styles of the input the user types in so the texts overlap properly. This PR adds a new `hintClasses` to complement `inputClasses`.

It uses the value of `inputClasses` when left `null`, but can take its own value if necessary to allow either:
- the styling of the hint on its own
- not styling the hint when stylint the inputClasses (by setting an empty string).


test/wdio.config.js
```diff
@@ -24,7 +24,7 @@ const capabilitiesLocal = [
   {
     browserName: 'chrome',
     'goog:chromeOptions': {
-      args: ['--headless=new'],
```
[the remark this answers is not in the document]
colinrotherham: Probably remove `args: []` entirely if you like

README.md
```diff
@@ -91,6 +91,10 @@ You can also import it using Sass:
 @import "accessible-autocomplete";
 ```
 
+> ![NOTE]
+>
+> If you add extra styles to adjust the `.autocomplete__input` elements, be aware that when using the `autoselect` option, the component overlays it above a second `.autocomplete__hint` input. That element displays the completions when what the user typed matches the start of a suggestion. It may need adjustment to account for the changes you may have made to `.autocomplete__input` so that the text of both align properly.
```
colinrotherham, line 96: Can simplify this a bit perhaps?

```suggestion
> When styling the `.autocomplete__input` element, be aware that the `autoselect` option will render a second `.autocomplete__hint` input element for suggestion text. Both elements should be styled to ensure suggestions exactly align with the typed input text.
```

test/wdio.config.js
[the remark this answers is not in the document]
romaricpascal: Oh no, that's a byproduct of me needing to see what happened in the tests. I'll need to revert it back, thanks for spotting it.

README.md
```diff
@@ -183,11 +187,21 @@ function suggest (query, populateResults) {
 
 ### Other options
 
-#### `inputClasses` (default: `''`)
+#### `inputClasses` (default: `null`)
+
+Type: `string`
```
colinrotherham, line 192:
```suggestion
Type: `string | null`
```

test/integration/index.js
```diff
@@ -222,6 +222,33 @@ const customTemplatesExample = () => {
   })
 }
 
+const classesPropsExamples = () => {
+  describe.only('classes properties', () => {
```
colinrotherham, line 226: Just got a stray `.only()` here

```suggestion
  describe('classes properties', () => {
```

README.md
```diff
@@ -183,11 +187,21 @@ function suggest (query, populateResults) {
...
+
+Adds custom html classes to the `input` elements of the component:
+- the one the user types in
+- the one used to display the completion hint when `autoselect` is set to `true` (see `hintClasses` hereafter)
```
colinrotherham, line 196: Think we should link between `hintClasses` and `inputClasses`?

```suggestion
Adds custom html classes to the generated `input` element.

If [`autoselect`](#autoselect) is set to `true`, the option [`hintClasses`](#hintClasses) can be configured separately or it will default to the `inputClasses` value.
```

README.md
```diff
@@ -183,11 +187,21 @@ function suggest (query, populateResults) {
...
+
+#### `hintClasses` (default: `null`)
 
 Type: `string`
 
-Adds custom html classes to the generated `input` element.
+Adds custom html classes to the hint element that appears when what the user typed matches the start of a suggestion (and `autoselect` is set to `true`).
+
+If `inputClasses` is set, it'll default to the value of `inputClasses`. You can set an empty string `''` to prevent the `inputClasses` being set on the hint as well.
```
colinrotherham, line 204: Likewise, linking from `hintClasses` to `inputClasses`?

```suggestion
Adds custom html classes to the additional `input` element used for suggestion hints.

If [`autoselect`](#autoselect) is set to `true`, the option [`inputClasses`](#inputClasses) will be used as the default value unless `hintClasses` is set to an empty string `''`.
```

examples/index.html
```diff
@@ -489,13 +512,42 @@ <h3>Translating texts</h3>
...
+      accessibleAutocomplete({
         element: element,
         id: id,
         source: countries,
         menuAttributes: {
           "aria-labelledby": id
         },
-        inputClasses: "govuk-input"
+        autoselect: true,
+        inputClasses: 'custom-input-class'
```
colinrotherham, line 535: Worth using an [`app-*` prefix](https://github.com/alphagov/govuk-frontend/blob/b44022f0ddb62268ee608c699670bc93529140fc/docs/contributing/coding-standards/css.md#govuk-namespacing) for a (future) `govuk-input app-input` suggestion?

```suggestion
        inputClasses: 'app-input'
```

colinrotherham (changes requested):
Looking good 🙌

Just the `.only()` to remove, the docs changes aren't blockers

test/integration/index.js
romaricpascal: Whoop, sorry for that 😬

README.md
```diff
@@ -191,17 +191,17 @@ function suggest (query, populateResults) {
...
+Adds custom html classes to the generated `input` element.
+
+If [`autoselect`](#autoselect) is set to `true`, the option [`hintClasses`](#hintClasses) can be configured separately or it will default to the `inputClasses` value.
 
 #### `hintClasses` (default: `null`)
 
 Type: `string`
 
-Adds custom html classes to the hint element that appears when what the user typed matches the start of a suggestion (and `autoselect` is set to `true`).
+Adds custom html classes to the additional `input` element used for suggestion hints.
```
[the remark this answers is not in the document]
romaricpascal: Sounds clearer. The `hint` naming confused me a little at start, so expliciting what it represents will be helpful. I'll amend the docs quickly and I _think_ that'll be all the feedback taken into account 💪🏻

colinrotherham (approved):
Looks good to me

Would be ace to see tests for:

1. Class only on input and NOT on the hint
2. Class only on hint and NOT on the input

Just to make sure `hintClasses: ''` etc are working

romaricpascal:
Yeah, I think it'd be good to look at how to make it easy to have that kind of tests without cluttering the examples:
- Karma tests don't have a system to wait for the UI to have changed before running assertions like WebDriver IO have
- More tests in WebDriver mean more examples on that example page, which is not ideal either. 😔

PR #650, opened 2024-01-26
Update ways of configuring menu attributes

romaricpascal:
Prevent the attributes set through `menuAttributes` from overriding any of the attributes set internally:
- `id` and `role` are dropped to limit risks to accessibility
- `className` and `class` are appended to the `class` attribute of the menu so it keeps the classes computed by the component

In addition to the update to `menuAttributes`, the PR also adds a `menuClasses`. Even if it doubles with with `menuAttributes.(class|className)`, it provides a API with a name consistent with `inputClasses` and `hintClasses` from #649 .


examples/index.html
```diff
@@ -57,6 +57,10 @@
       .custom-hint-class {
         box-shadow: inset 0 0 0 5px #44952e;
       }
+
+      .custom-menu-class {
+        box-shadow: 0 3px 0 0 #fd0, 3px 3px 0 0 #fd0, -3px 3px 0 0 #fd0;
```
romaricpascal, line 62: Bit wordy, but extends the yellow focus border to wrap around the menu, not just the input, for a nice visual change.

src/autocomplete.js
```diff
@@ -490,6 +487,29 @@ export default class Autocomplete extends Component {
       inputClassList.push(inputClasses)
     }
 
+    const menuClassName = `${cssNamespace}__menu`
+    const menuModifierDisplayMenu = `${menuClassName}--${displayMenu}`
+    const menuIsVisible = menuOpen || showNoOptionsFound
+    const menuModifierVisibility = `${menuClassName}--${(menuIsVisible) ? 'visible' : 'hidden'}`
+
+    const { classList, ...safeMenuAttributes } = sanitiseMenuAttributes(menuAttributes || {})
```
romaricpascal: Little safety with the `|| {}` in case someone passes `null` or `undefined`.

src/autocomplete.js
```diff
@@ -490,6 +487,36 @@ export default class Autocomplete extends Component {
       inputClassList.push(inputClasses)
     }
 
+    const menuClassName = `${cssNamespace}__menu`
+    const menuModifierDisplayMenu = `${menuClassName}--${displayMenu}`
+    const menuIsVisible = menuOpen || showNoOptionsFound
+    const menuModifierVisibility = `${menuClassName}--${(menuIsVisible) ? 'visible' : 'hidden'}`
+
+    const safeMenuAttributes = sanitiseMenuAttributes(menuAttributes || {})
```
colinrotherham, line 495: Instead of lots of new sanitisation code, can we overwrite role/id instead?

```js
menuAttributes.id = id
menuAttributes.role = role
```

src/autocomplete.js
```diff
@@ -490,6 +487,36 @@ export default class Autocomplete extends Component {
...
+
+    const menuClassList = [
+      menuClassName,
+      menuModifierDisplayMenu,
+      menuModifierVisibility
+    ]
+
+    // Both React and Preact support the `className` attribute for setting
+    // an element's class in JSX
+    if (safeMenuAttributes.className) {
```
colinrotherham, line 505: Over in https://github.com/alphagov/accessible-autocomplete/pull/649 we talked about empty string `''` values being valid

We should probably allow that here too?

```suggestion
    if ('className' in safeMenuAttributes) {
```

src/autocomplete.js
romaricpascal: That's a good idea for avoiding to `delete` the `id` and the `role`, didn't think of that. We could do the same with `class` and `className` (we'd need to set both, but I don't think that's a massive issue. Could try setting one to `null` depending on how things work with React/Preact). 

I'd still be keen to clone the object to avoid modifying the prop we're given, or do you think it's over the top?

colinrotherham: With object spread?

We had `{...menuAttributes}` on line 538 previously:

https://github.com/alphagov/accessible-autocomplete/blob/40350ae89328e813cc9437e4aac8d52ef813dd4a/src/autocomplete.js#L538C11-L538C30

Are you thinking to overwrite role/id with `{...menuAttributes, id, role}` instead?

src/autocomplete.js
romaricpascal: Empty strings would add extra spaces to the attribute for nothing, so I think we're better off excluding them. Good point about consistency, though, I'll have a check at how `inputClasses` handles them (`hintClasses` filters them out through the `filter(Boolean)`).

src/autocomplete.js
romaricpascal: Yeah, just with an object spread, so that when we set the `id` or `role`, it doesn't leak the modifications to the object that was passed as props directly. Thinking this'll end up with something like:

```js
// Compute the menu classes from the array
// including `menuAttributes.class` and `menuAttributes.className`
const menuClasses = ...

const computedMenuAttributes = {
  ...menuAttributes,
  // Override the attributes the component computes
  id: `${id}__listbox`,
  role: 'listbox',
  className: menuClasses.join(' ')
}

// Figure a way to delete the `class` property
delete computedMenuAttributes.class

// Then in the JSX, leaving the handler here for consistency with the other elements
// whose handlers are set  in JSX. But could equaly go in `computedMenuAttributes` as well.
<ul 
	onMouseLeave={(event) => this.handleListMouseLeave(event)}
	{...computedMenuAttributes} >
```

src/autocomplete.js
```diff
@@ -490,6 +487,44 @@ export default class Autocomplete extends Component {
...
+    }
+
+    const computedMenuAttributes = {
+      // Copy the attributes passed as props
+      ...menuAttributes,
+      // And add the values computed for the autocomplete
+      id: `${id}__listbox`,
+      role: 'listbox',
+      className: menuClassList.join(' '),
+      onMouseLeave: this.handleListMouseLeave
```
romaricpascal, line 520: Removed the wrapping arrow function, as the handler is bound in the constructor already, so that'll save re-creating listeners each time the menu is re-rendered.

README.md
```diff
@@ -209,6 +209,19 @@ Type: `Object`
...
+> [!NOTE]
+>
+> This option will not override the attributes computed by the component
+> - `id` and `role` are dropped to limit risks to accessibility
+> - `className` and `class` are appended to the `class` attribute of the menu so it keeps the classes computed by the component
+
+
+#### `menuClasses` (default: `null`)
+
+Type: `String`
```
colinrotherham, line 221:
```suggestion
Type: `string | null`
```

src/autocomplete.js
colinrotherham: >I'm not sure I follow there. Do you have any specifics in mind?

Nothing specific, just the PR description called out consistency as a feature:

>In addition to the update to `menuAttributes`, the PR also adds a `menuClasses`. Even if it doubles with with `menuAttributes.(class|className)`, it provides a consistent API with `inputClasses` and `hintClasses` from #649.

For example, where we destructure `this.props` into `menuAttributes`

Users would consistently think `{ menuAttributes: { className: '' } }` resets the class

romaricpascal: Oh, I see. I meant consistency in terms of naming, so I'll clarify that in the PR. If `hintClasses` clearing its default on `''` causes trouble with `menuClasses` or `menuAttributes.className` not doing the same, we can update when we get feedback 😊

README.md
```diff
@@ -209,6 +209,19 @@ Type: `Object`
 
 Sets html attributes and their values on the `menu`. Useful for adding `aria-labelledby` and setting to the value of the `id` attribute on your existing label, to provide context to an assistive technology user.
 
+> [!NOTE]
+>
+> This option will not override the attributes computed by the component
+> - `id` and `role` are dropped to limit risks to accessibility
+> - `className` and `class` are appended to the `class` attribute of the menu so it keeps the classes computed by the component
```
colinrotherham, line 216: I've had a go at re-wording this to include all three `id`, `role` and `onMouseLeave`

Good to explain and link to `menuClasses` too? Don't worry about explaining the `class`/`className` difference since we haven't already done so in the `dropdownArrow` docs

```suggestion
> To maintain assistive technology support, menu attributes `id`, `role` and `onMouseLeave` cannot be overridden using `menuAttributes`. Setting `className` will append to the component default and [`menuClasses`](#menuClasses) values.
```

src/autocomplete.js
```diff
@@ -490,6 +487,44 @@ export default class Autocomplete extends Component {
...
+      menuClassList.push(menuClasses)
+    }
+
+    if (menuAttributes?.className) {
+      menuClassList.push(menuAttributes.className)
+    }
+
+    if (menuAttributes?.class) {
+      menuClassList.push(menuAttributes.class)
+    }
```
colinrotherham, line 511: Looks like `class` takes precedence in Preact, rather than allowing both?

```suggestion
    if (menuAttributes?.class || menuAttributes?.className) {
      menuClassList.push(menuAttributes?.class || menuAttributes?.className)
    }
```

src/autocomplete.js
colinrotherham: Shall we do the same to `handleInputClick`?

src/autocomplete.js
romaricpascal: Yeah, that's a good shout. I'll amend 😊

romaricpascal: Added extra test cases while I was at it.

romaricpascal:
@colinrotherham Things should be all tidy. The first force-push has the interesting changes, the second being just the tidy up of a stray comma that `standard` wasn't happy with 😬

README.md
```diff
@@ -209,6 +209,17 @@ Type: `Object`
...
+> [!NOTE]
+>
+> To maintain assistive technology support, menu attributes `id`, `role` and `onMouseLeave` cannot be overridden using `menuAttributes`. Setting `className` will append to the component default and [`menuClasses`](#menuClasses) values.
+
+
+#### `menuClasses` (default: `null`)
+
+Type: `string | null`
+
+Adds custom html classes to the menu elements.
```
colinrotherham, line 221: We just say "menu" elsewhere but should we fix ~elements~ → element, or be specific like `menuAttributes`?

```suggestion
Adds custom html classes to the generated `ul` menu element.
```

romaricpascal: Updated both here and for `menuAttributes` 😊

colinrotherham (approved):

README.md
colinrotherham: Even better, thank you

PR #676, opened 2024-03-01
Fix binding of `handleInputClick`

romaricpascal:
Had missed that is was not bound like the other `handleInput...` functions when removing the wrapping arrow function in 25cc4a3.

This was displaying an error in the browser console as `this` was not defined when the method was called in response to clicks.

That error in the console seems the only effect of the function not being bound. Both with and without input, and with or without the `autoselect` option set, I did not notice any visual change in behaviour from the component.

We may want to review the purpose of that event handler, maybe the `handleInputFocus` one does enough? 🤔


romaricpascal:
Thinking we didn't catch it because we don't have anything failing tests when an error occurs inside the browser 😔 Opened https://github.com/alphagov/accessible-autocomplete/issues/679 to track adding this as it doesn't seem straightforward.

domoscargin (approved):

PR #678, opened 2024-03-01
Add notice about supported Preact versions in README

### LINKED ISSUE #667: Run test against latest version 10.x of Preact
## What

Run our Preact tests against an updated version of [the Preact example](https://github.com/alphagov/accessible-autocomplete/blob/main/examples/preact/index.html) loading Preact 10 (`https://unpkg.com/preact@10.19.6/dist/preact.js`) to check if the library is compatible.

Update the 'Using with Preact' section with the result, in a similar fashion that [the React section mentions its versions](https://github.com/alphagov/accessible-autocomplete?tab=readme-ov-file#react-versions).

## Why

Newer versions of Preact have been released since we last wrote the tests and it'll be good to let our users know whether we're compatible or not.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Tests have been run and README updated

romaricpascal:
#677 (not to be merged) provides the branch testing against Preact 10.19.6, which showed a single automated test failing in IE11.

This commits adds a notice to the README to clarify which versions of Preact were tested and the result, in a similar fashion to the section for React.

Closes #667

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,14 @@ preact.render(
 
 [Try out the Preact example!](https://alphagov.github.io/accessible-autocomplete/examples/preact/)
 
+#### Preact versions
+
+Preact v8.5.3 has been tested to work with the Accessible Autocomplete - although make sure to check out [documented issues](https://github.com/alphagov/accessible-autocomplete/issues).
+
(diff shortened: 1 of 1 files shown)


romaricpascal:
Converting back to draft while I re-check the results.

domoscargin (approved):
Looks sound!

PR #681, opened 2024-03-11
[DO NOT MERGE] Testing against React 16

romaricpascal:
Updates the React example to run against React 16 and check the compatibility of our code with this newer version of React, given we're using React 15.


romaricpascal:
Branch was temporary to run automated tests on CI

PR #683, opened 2024-03-11
[DO NOT MERGE] Testing against React 18

romaricpascal:
Updates the React example to run against React 18, including using [the new client API from React DOM](https://react.dev/blog/2022/03/29/react-v18#new-client-and-server-rendering-apis) ([`createRoot`](https://react.dev/reference/react-dom/client/createRoot#reference)), and check the compatibility of our code with this newer version of React, given we're using React 15.

PR #684, opened 2024-03-11
Update notice about supported React versions in README 

### LINKED ISSUE #668: Run test against newer versions of React
## What

Run our React tests against updated versions of [the React example](https://github.com/alphagov/accessible-autocomplete/blob/main/examples/react/index.html) loading the following versions of React:
- 16.0
- 17.0
- 18.0 (be aware of [changes of the API for rendering components on the page](https://react.dev/blog/2022/03/29/react-v18#new-client-and-server-rendering-apis))

Update the 'Using with React' section with the result, in a similar fashion that the React section mentions its versions.

## Why

Newer versions of React have been released since we last wrote the tests and it'll be good to let our users know whether we're compatible or not.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [x] We ran tests against React 16 - #681
- [x] We ran tests against React 17 - #682
- [x] We ran tests against React 18 - #683
- [ ] We updated the README - #684

romaricpascal:
Update the notice about React versions in the README with the results of:

- #681 
- #682 
- #683 

Tests didn't show any issue (accounting for React 18 dropping support for IE, meaning the tests fail for that branch), though it's only automated testing and some (very) light manual testing so we keep encouraging people to carry their own testing.

Closes #668

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,9 @@ ReactDOM.render(
 React v15.5.4 has been tested to work with the Accessible Autocomplete - although make sure to check
 out [documented issues](https://github.com/alphagov/accessible-autocomplete/issues).
 
-React v15.6.2 and 16.0 have been incompletely tested with the Accessible Autocomplete: while no undocumented issues were found, we recommend you carry out thorough testing if you wish to use these or later versions of React.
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
:+1:

Issue #565, opened 2022-09-26
aria-controls instead of aria-owns?


jeremyZX:
In reviewing the [WAI ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) and seeing to what extent this library follows that pattern, I noticed that the form input is using `aria-owns`, which appears to be a ARIA 1.0 recommendation. At the bottom of the page, there is a note:

> In ARIA 1.0, the combobox referenced its popup with [aria-owns](https://w3c.github.io/aria/#aria-owns) instead of [aria-controls](https://w3c.github.io/aria/#aria-controls). While user agents might support comboboxes with [aria-owns](https://w3c.github.io/aria/#aria-owns) for backwards compatibility with legacy content, it is strongly recommended that authors use [aria-controls](https://w3c.github.io/aria/#aria-controls).

Is there any possibility that you can update to use `aria-controls` instead of `aria-owns`, or a reason why one would not want to use aria-controls?

eugeniosegala:
Echoing @jeremyZX, I have seen this being flagged as a critical accessibility issue by AXE:

```
Fix any of the following: Required ARIA attribute not present: aria-controls
```

Any chances we could update this? Happy to create a PR but wanted to discuss here before.

36degrees:
We've [previously said this might be part of an intentional 'regression' to prefer the ARIA 1.0 pattern](https://github.com/alphagov/accessible-autocomplete/pull/386#issuecomment-550367646) due to better support from assistive technologies.

That was 4 years ago, so you'd hope that support for the 1.1 pattern had improved regardless, but we need to understand the nuance of this change and ensure it's thoroughly tested.

igneosaur:
This has also been flagged as part of our accessibility autit for a project in DWP. Would it be resonable to add both attributes? Then we'd be backward compatible and the component would no longer get flagged.

PR #725, opened 2024-08-14
Add `aria-controls` to aria props

owenatgov:
## What/Why
Adds `aria-controls` to the list of props in `ariaProps`.

Previously, we've made the decision to not include `aria-controls` in favour of `aria-owns` for better compatability with aria 1.0 and therefore better compatability with screen readers. [See Nick's comment](https://github.com/alphagov/accessible-autocomplete/pull/386#issuecomment-550367646) and [Ollie's comment](https://github.com/alphagov/accessible-autocomplete/issues/565#issuecomment-1900150271) on this topic.

This has been re-prompted by a DAC audit of the GOV.UK Design System website, where we use the autocomplete for our search. DAC advised the following:

>The combobox is missing the required aria-controls attribute. The lack of the aria-controls attribute on the combobox can be problematic as it does not associate the relationship between the combobox and its corresponding listbox. The aria-controls attribute is fundamental for providing an explicit relationship between these two elements, which is necessary for users who rely on assistive technologies such as screen readers. While there is an aria-owns attribute present and some user agents might support this, for backwards compatibility it is strongly recommended that the aria-controls attribute is implemented instead.

Following [testing with all current major screen readers](https://docs.google.com/spreadsheets/d/1k5TpqEg5LBQX14U-mAiD4_C8xdpW5pJ-W8kPpXuaa-w/edit?gid=0#gid=0) on `aria-owns`, `aria-controls` and both, we've found no difference between the 3 scenarios. Therefore, we can confidently add `aria-controls` to our aria props and remove `aria-owns`.

Resolves https://github.com/alphagov/govuk-design-system/issues/4004
Resolves https://github.com/alphagov/accessible-autocomplete/issues/565

## Notes
The testing isn't complete yet due to difficulties we've been having testing on mobile devices. Based on desktop testing, it's unlikely that anything's going to come out of mobile testing, however for completeness we'll aim to resolve mobile testing before merging.


owenatgov:
I'm going to have to close this and reopen it as a different PR in order to trigger a netlify deploy preview for testing

PR #729, opened 2024-08-16
Move `aria-labelledby` into computed menu props

owenatgov:
## What/Why
The examples include `aria-labelledby`'s passed to the autocomplete via `menuAttributes`. This change moves `aria-labelledby` into the computed menu attributes so that it's provided by default.

This was prompted by a recent DAC audit of the GOV.UK Design System website which noted that the `aria-labelledby` is absent from our site search. We could solve this fairly quickly by just adding `menuAttributes` to the site search autocomplete, but solving it upstream is the more holistic option. This also means all users of the autocomplete don't need to add their own `aria-labelledby`.

Resolves https://github.com/alphagov/govuk-design-system/issues/4005

## Notes
We could also use `aria-label` instead of `aria-labelledby`, however I'm not sure what `aria-label` could be. We could potentially explore a dynamic `aria-label` but this could be challenging when trying to take internationalisation into account. In my opinion, `aria-labelledby` is the more managable option.


src/autocomplete.js
```diff
@@ -508,13 +508,14 @@ export default class Autocomplete extends Component {
     }
 
     const computedMenuAttributes = {
-      // Copy the attributes passed as props
-      ...menuAttributes,
       // And add the values computed for the autocomplete
+      ...menuAttributes,
       id: `${id}__listbox`,
       role: 'listbox',
+      'aria-labelledby': id,
```
romaricpascal, line 515: **issue** This would be a breaking change for people who may have set the ID to something outside the component. Unlike `id` and `role`, which we need to be in control of, we can happily let people use anything else to label the menu. It's a cheap change as we only need to re-order the properties of the object to avoid the breaking change while setting a sensible default that automatically labels the menu:

```js
{
  'aria-labelledby': id,
  ...menuAttributes,
  id: `${id}__listbox`,
  /* ... */
}

File: src/autocomplete.js [hunk 2]
```diff
@@ -508,13 +508,14 @@ export default class Autocomplete extends Component {
...
       className: menuClassList.join(' '),
       onMouseLeave: this.handleListMouseLeave
+      // Copy the attributes passed as props
```
romaricpascal, line 518: **nitpick** I'm not sure this comment ended up in the right line, it should be before `...menuAttributes`, I think.

romaricpascal (commented):
Neat, it does clean up all these examples with the `menuAttributes` 🙌🏻 It's bringing a tiny breaking change in its current state, though. Will other PRs bring breaking changes for the accessible-autocomplete? If not, I've proposed a fix to set the default `aria-labelledby` non breaking.

src/autocomplete.js
owenatgov: You are correct. This is a carryover from when I was messing with that object and thought it might be better for `...menuAttributes` to be declared after the other attributes as an override, which we don't wanna do. Was a symptom of me initially misunderstanding how the computed attributes worked. Will fix!

test/functional/index.js
```diff
@@ -111,15 +111,13 @@ describe('Autocomplete', () => {
...
           const wrapperElement = scratch.getElementsByClassName('autocomplete__wrapper')[0]
           const dropdownElement = wrapperElement.getElementsByTagName('ul')[0]
-          const inputElement = wrapperElement.getElementsByTagName('input')[0]
 
-          expect(dropdownElement.getAttribute('aria-labelledby')).to.equal('autocomplete-default')
-          expect(inputElement.getAttribute('id')).to.equal('autocomplete-default')
+          expect(dropdownElement.getAttribute('data-test')).to.equal('test')
         })
 
         describe('attributes computed by the component', () => {
```
romaricpascal, line 123: **issue** I think we'd still want a test in here to check that it defaults the `aria-labelledby` to the `id` (and ideally one that it can be overriden, but that one's not super critical).

owenatgov: I've added a test for both checking that the labelledby is the id by default and that it gets overriden by `menuAttributes`

src/autocomplete.js
```diff
@@ -508,6 +508,8 @@ export default class Autocomplete extends Component {
     }
 
     const computedMenuAttributes = {
+      // set aria-labelledby first so that users can override it with menuAttributes
```
romaricpascal, line 511: Good shout adding a comment 😍

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #727: Add `aria-controls` to aria props](https://github.com/alphagov/accessible-autocomplete/pull/727)
```
romaricpascal, line 7: I think the PR is missing a little rebase. Would be good to add an entry for this fix in the changelog as well 😊

romaricpascal (approved):

selfthinker:
From the description it wasn't clear to me what this PR did, specifically how it fixed the issue raised.
Now that I've tested it, I found the fix is to label the listbox with "[search term] list".

It might have been better to give it the same name independent of the search term, like "results list".
Although there might also be benefits in labelling it differently every time to make clear it's a different list.
Altogether I think it works well enough as it is. 👍

PR #753, opened 2024-10-08
Add visual cue for when a dialog is available

owenatgov:
## Change

Adds visual cues to the component so that users who are zoomed into the input know that there is a dialog below that they may be able to interact with. When a dialog is displayed, the input will be given a 'bottomless' class to remove it's bottom border, making it visually flow into the menu and drawing the eye down.

This also gates the change behind the `useBottomlessInput` attribute so that users using the `cssNamespace` attribute doesn't get served useless CSS classes on render.

More detail in https://github.com/alphagov/govuk-design-system/issues/4015

### Before/without attribute
Autocomplete with open menu before change

### After
Autocomplete with open menu after change


src/autocomplete.css
```diff
@@ -17,28 +17,40 @@
...
+  box-shadow: inset 2px 2px 0 0, inset -2px 0 0 0;
+}
+
+.autocomplete__dropdown-arrow-down-wrapper {
+  display: none;
+}
+
+.autocomplete__dropdown-arrow-down-wrapper--visible {
+  display: block;
+}
```
romaricpascal, line 52: **question** Given the `dropdownArrowFactory` accepts a `className`, what's the reason for adding an extra wrapper element with its own class rather than using a `className` with two class names separated by a space?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #753: Add visual cue for when a dialog is available](https://github.com/alphagov/accessible-autocomplete/pull/753)
```
romaricpascal, line 7: **question** Could this be big enough a change that it warrants a 'New feature' entry with a little more noise?

test/integration/index.js
```diff
@@ -21,7 +21,7 @@ const basicExample = () => {
 
     beforeEach(async () => {
       $input = await $(input)
-      $menu = await $(`${input} + ul`)
+      $menu = await $(`${input} ~ ul`)
```
romaricpascal, line 24: That'll make our tests more resilient to markup changes 🙌🏻

romaricpascal (commented):
Added a couple of questions on the code changes, but couldn't check how things were working as Netlify seems to be struggling to deploy. From the screenshots, the arrow looks quite high inside the input box, though, is this intended?

src/autocomplete.js
```diff
@@ -461,29 +463,32 @@ export default class Autocomplete extends Component {
       'aria-autocomplete': (this.hasAutoselect()) ? 'both' : 'list'
     }
 
-    let dropdownArrow
-
-    // we only need a dropdown arrow if showAllValues is set to a truthy value
-    if (showAllValues) {
-      dropdownArrow = dropdownArrowFactory({ className: dropdownArrowClassName })
+    let dropdownArrow = dropdownArrowFactory({
+      wrapperClassName: dropdownArrowWrapperClassName,
```
romaricpascal, line 467: **issue** I think this is a breaking change. The current `dropdownArrow` option only expects a `className`, meaning existing implementations will ignore the new `wrapperClassName` and potentially show an arrow when they're not meant to.

src/autocomplete.css
romaricpascal: See https://github.com/alphagov/accessible-autocomplete/pull/753/files#r1791991077 as well. It'd be neat if we could only use the existing `className` prop to control the arrow's visibility.

Another option could be to set a class on a wrapping element and target the `dropdownArrowClassName` we set to control visibility.

owenatgov:
@edwardhorsford This is useful context, cheers. To clarify, the arrow is only shown when there are results. I've tried to adjust the content of the PR to make that clearer.

edwardhorsford:
> @edwardhorsford This is useful context, cheers. To clarify, the arrow is only shown when there are results. I've tried to adjust the content of the PR to make that clearer.

Thanks, I understand that - I'd still be cautious with this change and encourage usability testing of it first to make sure the downsides I mentioned don't occur.

owenatgov:
After discussing this again with the team, we're going to remove the arrow following @edwardhorsford's comments about prior research. We think the 'bottomless' effect is good enough (pending testing by @selfthinker). I'm going to remove the arrow and adjust the description of this PR.

edwardhorsford:
Thanks @owenatgov - I didn't comment on the line but it looks like a nice visual touch! 

It may well be there's no impact from the arrow, or another visual indicator could be added that doesn't look like a arrow / select.

owenatgov:
I've added an attribute to this PR: `useBottomlessInput`. This gates this process behind an attribute to prevent us service useless CSS classes to users of `cssNamespace`. This is following suggestions by Romaric on this PR and discussions off github that this could be a breaking change.

Now that I've done it I'm not sure how necessary this is. The component won't stop working, it'll just not use the visual enhancement and we service them extra html in the form of additional CSS classes. I'm not sure this is breaking necessarily. The 'breaking' point felt much more relevant when we still had the arrow since the way I'd done it made it more likely to be a breaking change. @romaricpascal Whatcha reckon?

selfthinker:
I have tested the old and new version in Windows High Contrast Mode (option 1, 2 and white), Firefox colour changes (a dark and a light colour scheme) and the Chrome extension 'High Contrast'.
I have uploaded screenshots of everything to [this folder](https://drive.google.com/drive/folders/1ApXe_x9eN-BUkl06A1Ni16EZO-7pSXOp?usp=drive_link) (which can only be viewed internally).

To summarise those:
I guess the difference is clear enough when none of the results has been selected yet. But when the first item has been selected (as is the case automatically on the Design System website), the difference is much less visible.
In Firefox the dropdown never looks like an extension of the text field because their background colours are different.

The dropdown box never aligns with the text field. That has nothing to do with this change, though.
It's also interesting that the highlighted item always has blue as a background colour. While that also doesn't have anything to do with this change, it might be more important to help with the perception that the dropdown extends the text field. There might be a problem with how we set the colour. Could we maybe use the [system colour 'Highlight'](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#highlight) in forced colour modes?

owenatgov:
Following a chat off-github with @romaricpascal I'm going to close this in favour of https://github.com/alphagov/govuk-design-system/pull/4220. Reasons:

- We expect teams to mostly do their own styling of their autocompletes and keep the base level CSS light
- As I've found with https://github.com/alphagov/govuk-design-system/pull/4220, this can be done 100% with css and we'd have to reapply this styling ourselves anyway

If this turns out to be useful for users who aren't styling their own autocompletes, we may reopen or restart this.

PR #768, opened 2024-12-11
Add automatic redirects to examples for netlify previews

owenatgov:
What it says on the tin. Adds a netlify.toml that only contains the redirect from "/" to "/examples". Used only for netlify previews right now so that folks not in the know aren't wondering why the root of the preview doesn't show anything.


domoscargin (approved):
Noice!

Issue #752, opened 2024-10-07
value-error missing from aria-describedBy


jackheslop96:
In our most recent DAC audit, a AAA was reported relating to:

`aria-describedBy=value__assisstiveHint` on the accessible autocomplete input.

It was suggested we add:

`aria-describedBy=value__assisstiveHint value-error`

Issue #589, opened 2023-04-20
enhanceSelectElement does not correctly link to existing hint or error message


edwardhorsford:
When using the `enhanceSelectElement`, an autocomplete may have a hint or error message, but unfortunately these are *not* linked with `aria-describedBy` from the input - so assistive technologies cannot get them.

The fix for this would be for the autocomplete code to grab the existing ids from `aria-describedby` from the select it's enhancing, and include those in addition to the instructions span it adds.

In the mean time, a fix is that the instructions added by the autocomplete are configurable by teams. We can make sure the instructions *include* any existing hint / error message.

When initialising an a given select as an autocomplete:

- Get the ids mentioned in the select `aria-describedby` - there can be more than one when the field is in error.
- Get the text content of those elements.
- Concat the text content together.
- Append with any instructions for AT for how to use the autocomplete
- provide this as `tAssistiveHint` when initialising the autocomplete.

edwardhorsford:
An alternate solution would be to use js to look up the original contents of `aria-describedBy` in the original select, then prepend these to the generated input.

Issue #586, opened 2023-03-16
Using hint with enhanceSelectElement


ShahSheel:
I have read through the documentation and using `hint`  and there is no option to specify a hint using a `enhanceSelectElement`

Eg:

```javascript
accessibleAutocomplete.enhanceSelectElement({defaultValue: '',
  selectElement: document.querySelector('#foobar'),
  hint: 'Begin searching. Use arrow keys to navigate options. Press Enter to select.' //Doesn't work
});
```

Using a `<label>` with `ariaDescribeBy` property will not work as the current functionality seems to prevent screen readers announcing this 

```html
<label class="form-label" for="foobar">Foobar</label> 
<span class="form-hint" id="hint-instructions">Begin searching. Use arrow keys to navigate options. Press Enter to select.</span> 
<select aria-describedby="hint-instructions" aria-label="Select a foobar" class="form-control" id="foobar" name="foobar" role="combobox">
    <option value="1">1</option>
</select>
```

If removing the JS function, then will announce properly as normal but the issue is it needs to use `enhanceSelectElement`.

ShahSheel:
Any insight into how to tackle this issue? Were blocked on this from trying to reach compliance at the moment

edwardhorsford:
We've discussed this a bit on slack, but I'm adding an answer here in case others find this issue.

I think the reason there's no `hint` option is because `enhanceSelectElement` is meant for enhancing existing markup. If the existing select you're enhancing has a hint, then it will get copied to the new component. 

From our discussions, it sounds like you want the hint that the new component has to be different than the select hint - this isn't supported unfortunately. But you can write your own jus to modify hints however you like.

I think the two ways of doing this would be:
* Before initialising the autocomplete, use js to loop through all selects of type autocomplete on the page, and update their hints with whatever you want.
* After initialising the autocompletes, loop through all autocompletes on the page, and update their hints with whatever you want.

edwardhorsford:
As an aside: I believe the autocomplete already includes instructions for how to use it for screen reader users. So you shouldn't need to add your own or show them visibily.

It's possible to modify these instructions if you like - it's the config option `tAssistiveHint`.

ShahSheel:
Reverted to use 2.0.3 as mentioned in slack as 2.0.4 seems to break. 

However this still doesn't get announced with assistive technologies. 

Tested on Edge + Chrome using built in MacOS Voice Over.

Seems to be related to: #498 , #567

[...]

PR #769, opened 2024-12-16
Propose accessibility fixes for enhanceSelectElement

oscarduignan:
These proposed changes would fix the following two issues

https://github.com/alphagov/accessible-autocomplete/issues/589
https://github.com/alphagov/accessible-autocomplete/issues/752

these are something that hmrc services get asked to fix in accessibility audits so the services have to patch their usages with some js, [we've been investigating how to roll the fixes into hmrc-frontend](https://github.com/hmrc/hmrc-frontend/pull/419) to save services having to add it themselves, but it seems like the fixes would be quite simple to make here so thought we'd propose them upstream just in case it's useful


ShahSheel:
Noticed this is in draft, I provided a custom approach here, hope it helps?

https://github.com/alphagov/accessible-autocomplete/issues/586

PR #794, opened 2026-08-17
Ensure hidden suffix for iOS is not visible when CSP block inline styles

### LINKED ISSUE #556: inline styles on iOS posinset won't work without unsafe-inline in CSP
Related issue to https://github.com/alphagov/accessible-autocomplete/issues/398, the [inline styling](https://github.com/alphagov/accessible-autocomplete/blob/d892838bd2f634a4261a121eaa135cf1b502fc6b/src/autocomplete.js#L527-L531) applied to visually-hide the span for count/index announcements is ignored on sites that have a strict CSP that doesn't allow inline styles, meaning that you see the count/indexes in the results:

Screenshot 2022-05-12 at 10 58 42
 
Can this be moved to CSS class instead?

barnabycollins:
Resolves #556; related to #398

The suffix discussed here is an additional span element, added only in iOS devices, which specified the position of each dropdown entry in the overall list. This span should be hidden so that users do not see it. This way, only a screen reader would see it. This is presumably a workaround for a deficiency in the iOS screen reader that means it will not give numbers for dropdown / list elements when other screen readers do.

Previously, the style ruleset hiding the screen reader suffix was only written inline which caused CSP issues. The code to hide the element would be disabled in any CSP-enabled site that hadn't set the proper CSP hash, meaning iOS users would see visible numbering, like (1 of 200), against each entry in the dropdown list.

This change copies the style to a new class, specified in the package CSS, and applies that class to the span, so that it will be visually hidden without requiring special CSP headers to be set on sites using the component. The inline styles are retained for redundancy on sites that have the proper CSP config, in case the CSS file fails to load.


Spencerooni:
We literally just encountered this issue yesterday and were wondering why the span was visible and what was going on... 🙏

src/status.js
```diff
@@ -89,7 +89,7 @@ export default class Status extends Component {
 
     return (
       <div
-        className={className}
+        className={'autocomplete__visually-hidden' + className}
```
NickColley:
```suggestion
        className={'autocomplete__visually-hidden ' + className}
```

Avoid risk of className being set and it turning into "autocomplete__visually-hiddenapp-classname" etc.

src/autocomplete.js
```diff
@@ -572,9 +572,8 @@ export default class Autocomplete extends Component {
             const optionModifierFocused = showFocused && hovered === null ? ` ${optionClassName}--focused` : ''
             const optionModifierOdd = (index % 2) ? ` ${optionClassName}--odd` : ''
             const iosPosinsetHtml = (isIosDevice())
-              ? `<span id=${id}__option-suffix--${index} style="border:0;clip:rect(0 0 0 0);height:1px;` +
-                'marginBottom:-1px;marginRight:-1px;overflow:hidden;padding:0;position:absolute;' +
-                'whiteSpace:nowrap;width:1px">' + ` ${index + 1} of ${options.length}</span>`
+              ? `<span id=${id}__option-suffix--${index} class=autocomplete__visually-hidden>` +
```
NickColley: I think we should mirror what you've done in the other instance of this class where, for now, we have the inline style and the class name so we can release a non-breaking release, then we raise a GitHub issue to remove the inline styles if we do a breaking change...

NickColley: We also need to make sure the HTML is `class="autocomplete__visually-hidden"`

NickColley (changes requested):

barnabycollins:
Update changelog also

NickColley (approved):
Thanks for putting this together @barnabycollins.

Since it's outside our team our policy is to get a second review so will ping the team and when someone has time they'll give a second look.

Then we'll try to release this as part of 3.0.2 when we can.

src/autocomplete.js
```diff
@@ -572,9 +572,10 @@ export default class Autocomplete extends Component {
             const optionModifierFocused = showFocused && hovered === null ? ` ${optionClassName}--focused` : ''
             const optionModifierOdd = (index % 2) ? ` ${optionClassName}--odd` : ''
             const iosPosinsetHtml = (isIosDevice())
-              ? `<span id=${id}__option-suffix--${index} style="border:0;clip:rect(0 0 0 0);height:1px;` +
-                'marginBottom:-1px;marginRight:-1px;overflow:hidden;padding:0;position:absolute;' +
-                'whiteSpace:nowrap;width:1px">' + ` ${index + 1} of ${options.length}</span>`
+              ? `<span id=${id}__option-suffix--${index} class="autocomplete__visually-hidden" ` +
```
romaricpascal: **issue** The `autocomplete` part of the classes is built from the `cssNamespace` variable so projects [can use their own namespace should they need to](https://github.com/alphagov/accessible-autocomplete/#cssnamespace-default-autocomplete).

```suggestion
              ? `<span id=${id}__option-suffix--${index} class="${cssNamespace}__visually-hidden" ` +
```

src/status.js
```diff
@@ -89,7 +89,7 @@ export default class Status extends Component {
 
     return (
       <div
-        className={className}
+        className={'autocomplete__visually-hidden ' + className}
```
romaricpascal: **issue** In a similar way as inside `autocomplete.js`, the `autocomplete` part of `autocomplete__visually-hidden` needs to use `cssNamespace`.

Simplest would be to pass both `statusClassName` and `${cssNamespace}__visually-hidden` when [setting the `Status` `className` prop](https://github.com/govuk-one-login/accessible-autocomplete/blob/21ecaf92088e4950fa9897a92c15903eebdd7589/src/autocomplete.js#L542) with the following, even though it separates the `style` attribute and the class that duplicates its CSS properties (we could always have the Status receive a `style` prop if we want to keep them close).

```
className=`${statusClassName} ${cssNamespace}__visually-hidden`
```

romaricpascal (changes requested):
Cheers for implementing this fix 😊 

It needs a little update before we can merge it to make sure it responds appropriately to the `cssNamespace` option which allows users to change the `autocomplete` part of the class names. 

We should probably add a little note in the CHANGELOG as well to explain to users that if they use `cssNamespace`, they'll need to make sure they implement the `__visually-hidden` class in their CSS to benefit from the fix.

NickColley:
Good suggestions @romaricpascal

I had a go (before realising I cant push to One Login branches :P) and I think there's a place where we can construct a variable for this new classname as well so here's the diff @barnabycollins hopefully helps:

https://github.com/alphagov/accessible-autocomplete/commit/b4dc3d9103eefd9de8f5fbd147265b94061b4b31

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #794: Copy iOS suffix inline styles to the stylesheet, to allow the component to be rendered correctly without specific CSP exceptions](https://github.com/alphagov/accessible-autocomplete/pull/794)
```
romaricpascal: I think we need a little more than a link to the pull request to guide people on how to update and maybe give it a little more visibility than "Fixes".

```suggestion
### Recommended change

#### Ensure visually hidden elements are not visible when Content Security Policies block inline styles

We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden element remaining visible when Content Security Policies block inline styles. 

If you copied the `dist/accessible-autocomplete.min.css` stylesheet from the package into your project, copy the stylesheet from this new version.

If you use the [`cssNamespace` option](https://github.com/alphagov/accessible-autocomplete#cssnamespace-default-autocomplete) to customise the class names applied by the accessible autocomplete, make sure your stylesheets contain a `<NAMESPACE>__visually-hidden` class with the [same declarations as `autocomplete__visually-hidden`](https://github.com/alphagov/accessible-autocomplete/blob/main/src/autocomplete.css#L169-180).

We made this change in [pull request #794: Ensure hidden suffix for iOS is not visible when CSP block inline styles](https://github.com/alphagov/accessible-autocomplete/pull/794)
```

romaricpascal (approved):
Thanks for setting up that `visuallyHiddenClassName` variable @NickColley. Code looks good to merge. I think we need to work out the CHANGELOG. I've made a proposition for some more content. @seaemsi, not urgent, but it could likely use some technical writing input 😅 

I'd propose to rename the PR to "Ensure hidden suffix for iOS is not visible when CSP block inline styles" as well to focus on the problem it solves rather than the solution.

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
 
+### Recommended change
+
+#### Ensure visually hidden elements are not visible when Content Security Policies block inline styles
+
+We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden element remaining visible when Content Security Policies block inline styles. 
+
+If you copied the `dist/accessible-autocomplete.min.css` stylesheet from the package into your project, copy the stylesheet from this new version.
+
+If you use the `cssNamespace` option to customised the class names applied by the accessible autocomplete, make sure your stylesheets contain an [equivalent `<NAMESPACE>__visually-hidden` class](https://github.com/alphagov/accessible-autocomplete/blob/main/src/autocomplete.css#L169-180).
```
NickColley:
```suggestion
```
Suggest removing this since we cant link to the correct line and it's a power-user sorta thing so they'll figure it out, move the link to the dist bit above instead

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Recommended change
+
+#### Ensure visually hidden elements are not visible when Content Security Policies block inline styles
+
+We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden element remaining visible when Content Security Policies block inline styles. 
+
+If you copied the `dist/accessible-autocomplete.min.css` stylesheet from the package into your project, copy the stylesheet from this new version.
```
NickColley:
```suggestion
If you copied the [`dist/accessible-autocomplete.min.css` stylesheet](https://github.com/alphagov/accessible-autocomplete/blob/main/src/autocomplete.css) from the package into your project, copy the stylesheet from this new version.
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Recommended change
+
+#### Ensure visually hidden elements are not visible when Content Security Policies block inline styles
```
seaemsi:
```suggestion
#### Make sure visually hidden elements are not visible when Content Security Policies block inline styles
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+
+We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden element remaining visible when Content Security Policies block inline styles. 
+
+If you copied the [`dist/accessible-autocomplete.min.css` stylesheet](https://github.com/alphagov/accessible-autocomplete/blob/v3.0.2/src/autocomplete.css) from the package into your project, copy the stylesheet from this new version.
```
seaemsi:
```suggestion
Make sure you're using the latest version of the [`dist/accessible-autocomplete.min.css` stylesheet](https://github.com/alphagov/accessible-autocomplete/blob/v3.0.2/src/autocomplete.css), which includes the `autocomplete__visually-hidden` class.
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+
+We made this change in [Pull request #794: Ensure hidden suffix for iOS is not visible when CSP block inline styles](https://github.com/alphagov/accessible-autocomplete/pull/794)
```
seaemsi:
```suggestion
We made this change in [pull request #794: Ensure hidden suffix for iOS is not visible when CSP block inline styles](https://github.com/alphagov/accessible-autocomplete/pull/794)
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Recommended change
+
+#### Ensure visually hidden elements are not visible when Content Security Policies block inline styles
+
+We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden element remaining visible when Content Security Policies block inline styles.
```
seaemsi:
```suggestion
We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden elements remaining visible when Content Security Policies block inline styles. 
```

CHANGELOG.md
NickColley:
```suggestion
Make sure you're using the latest version of the [`dist/accessible-autocomplete.min.css` stylesheet](https://github.com/alphagov/accessible-autocomplete/blob/v3.0.2/src/autocomplete.css), which includes the `autocomplete__visually-hidden` class.
```

From @seaemsi

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+### Recommended change
+
+#### Make sure visually hidden elements are not visible when Content Security Policies block inline styles
+
+We've added a new `autocomplete__visually-hidden` class to the component's stylesheet to fix visually hidden elements remaining visible when Content Security Policies block inline styles. 
+
+Make sure you're using the latest version of the [`dist/accessible-autocomplete.min.css` stylesheet](https://github.com/alphagov/accessible-autocomplete/blob/v3.0.2/src/autocomplete.css), which includes the `autocomplete__visually-hidden` class.
+
+We made this change in [pull request #794: Ensure hidden suffix for iOS is not visible when CSP block inline styles](https://github.com/alphagov/accessible-autocomplete/pull/794)
```
NickColley:
```suggestion
We made this change in [pull request #794: Ensure hidden suffix for iOS is not visible when CSP block inline styles](https://github.com/alphagov/accessible-autocomplete/pull/794).
```

seaemsi (approved):

Issue #730, opened 2024-08-20
Handle Sauce Connect 4's end of life


romaricpascal:
## What
Check that our CI will still be able

## Why

Sauce Connect 4 [reaches end of life at the end of December 2024](https://docs.saucelabs.com/secure-connections/sauce-connect/lifecycle/). In CI, our [WebDriver uses Sauce Connect under the hood](https://github.com/alphagov/accessible-autocomplete/blob/main/test/wdio.config.js#L110) to connect to SauceLabs and run against[ a variety of browsers](https://github.com/alphagov/accessible-autocomplete/blob/main/test/wdio.config.js#L110). 

This feature is provided to WebDriver [through the `saucelabs` package](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-sauce-service/package.json#L39C6-L39C17), the official node client for SauceLabs.

NickColley:
I believe that they're just not running anymore since I just checked a run and it didnt show up on the dashboard.

romaricpascal:
Had a little dig into this. So far, what I found:
- [Webdriver.io upgraded to saucelabs 9.0.1 to support Sauce Connect 5](https://github.com/webdriverio/webdriverio/commit/590bee4ca6ce25be1f907b45f47d541f7fc2945a)
- GitHub's UI is misleading, this wasn't shipped in 8.48.1 (there's no 8.48.1 for `@wdio/sauce-service`), but in Webdriver.io's next breaking release, as the feature [wasn't meant to be backported in v8](https://github.com/webdriverio/webdriverio/pull/14130)
- Upgrading to Webdriver 9.0 will require some hands-on work as they r[emoved the `devtool` protocol](https://webdriver.io/blog/2024/08/15/webdriverio-v9-release#removal-of-devtools-and-wdiodevtools-service-packages) that [our tests use when run locally](https://github.com/alphagov/accessible-autocomplete/blob/00979884732ce00f9dfa0b2fb5f37d9d837048dd/test/wdio.config.js#L80)

PR #796, opened 2026-09-08
[Temporary] Copy iOS suffix inline styles to component stylesheet

NickColley:
Raising to have Saucelabs do a run since they're skipped on forks...


NickColley:
Closing for now, I have done manual testing to verify the iPhone on Browserstack shows what we'd expect see #730 for more details on Saucelabs not running

